PR #60, opened 2023-09-21
build(c): add missing change from #52

chokobole:
# Description

#52 moves MSM / MSM implementations into each curve generated msm files. Therefore, `libtachyon.so` should depend on this!


TomTaehoonKim (approved):
LGTM

PR #63, opened 2023-09-25
fix(math): fix bit iteration in case of little endian

TomTaehoonKim:
# Description

While implementing Radix2 Evaluation Domain, test for evaluation on vanishing polynomials could not pass. Especially, cases with degree 1 polynomials were a problem.

@chokobole helped out fixing a bug in BitIterator in the case of little endian.

Unit test for sparse polynomial was missing a case with degree 1 polynomial, so I added one. Plus, I added another case with a polynomial with a degree 1 term.


chokobole (approved):
LGTM

Insun35 (approved):
LGTM

PR #67, opened 2023-09-26
perf: skip some swcurve operations

chokobole:
# Description

`CurveConfig::A()` and `CurveConfig::B()` are used to check whether point is on a given curve and group operation like addition and doubling. `IsOnCurve()` is not a critical path in terms of performance. But addition and doubling is. But when seeing `CurveConfig::A()` for bn254 and bls12_381, they are zero! We can skip some of instruction if we can determine at compile time. Plus, using `ScalarMul()` in semigroups, it needs to iterate all the bits, if you know `A()` in advance, you can make a fater multiplication code from generator.

After all these, I found that there's no point to differentiate `SWCurveGpu` from `SWCurve`. So this commit also incorporates those twos.


TomTaehoonKim (approved):
LGTM

Insun35 (approved):
LGTM

PR #68, opened 2023-09-27
fix(math): fix `IsOne` for sparse univariate polynomials

TomTaehoonKim:
# Description

Before this PR, `IsOne` did not check whether the degree is 0 or not. This PR fixes it.


chokobole (approved):
LGTM

PR #66, opened 2023-09-26
feat(math): add GLV parameters to bn254 g1

chokobole:
# Description

This PR adde `GLV` parameters to bn254 G1. (also includes some minor refactoring on glv test code)

PR #69, opened 2023-09-27
feat(math): implement `G2`

chokobole:
# Description

This implements `G2` for bn254 and bls12_381.

In `G2`, the static fields for SWCurve, are not able to be defined as `BigInt<N>` any longer, which can be marked with `constexpr`. At this moment, we have to express them with `BaseField` and `ScalarField`, which is on top of `PrimeField<T>` and one of them, `PrimeFieldGmp<T>` is not `constexpr` one. This results in a huge change!

Plus, I found that https://github.com/kroma-network/tachyon/pull/66 contains a set of wrong parameters and I don't
know why the CI passes. See
https://github.com/arkworks-rs/curves/blob/0a64024ebc5dc17364e849e32601883d7d12aa84/bn254/src/curves/g1.rs#L60-L65

So this commit also includes a fix for GLV, which is, it writes GLV coefficients as a `mpz_class` instead of `ScalarField`.


Insun35 (approved):
LGTM

Leegwangwoon (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #85, opened 2023-10-10
ci: fix cpplint script to run only in PR

Insun35:
# Description

Modify the script so that it runs only when there is a PR event.
There is a problem with the cpplint script when the approved PR is merged into the main branch.


chokobole (approved):
LGTM

PR #87, opened 2023-10-13
refac: enable omitting ostream operator

chokobole:
# Description

Now the google logging knows how to stream the value if the type of value supports `ToString()`! (Also this PR includes cpplint fix commit on the codes that we manage directly)

```c++
class F {
  std::string ToString() const { return "F()"; }
};

F f;
LOG(INFO) << f;
```

- If `operator<<()` is implemented for type T and is incorrectly streamed as boolean(such as function and function pointer), then glog stringify it with `std::addressof()`.
- If `ToString()` is implemented for type T, then glog uses `ToString()`.
- Else, `operator<<()` should be implemented for type T.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,7 +17,7 @@ IncludeBlocks: Regroup
 IncludeCategories:
   # C system headers.  The header_dependency_test.py contains a copy of this
   # list; be sure to update that test anytime this list changes.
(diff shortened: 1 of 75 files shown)


TomTaehoonKim (approved):
LGTM! Can you fix the CI failing?

Insun35 (approved):
LGTM

PR #88, opened 2023-10-13
feat: enable customized group inverse

chokobole:
# Description

Previously, `Negative()` and `Inverse()` depends on `NegInPlace()` and `InverseInPlace()` internally. But there's a case that those functions require the return type to be different. In this case, the derived class needs to define its own `Negative()` and `Inverse()`.

### DIFF
--- a/tachyon/math/base/groups.h
+++ b/tachyon/math/base/groups.h
@@ -48,7 +48,7 @@ class MultiplicativeGroup : public MultiplicativeSemigroup<G> {
   template <
       typename G2,
       std::enable_if_t<internal::SupportsDivInPlace<G, G2>::value>* = nullptr>
-  constexpr auto& operator/=(const G2& other) {
+  constexpr G& operator/=(const G2& other) {
     G* g = static_cast<G*>(this);
     return g->DivInPlace(other);
   }
@@ -58,7 +58,7 @@ class MultiplicativeGroup : public MultiplicativeSemigroup<G> {
       typename G2,
       std::enable_if_t<!internal::SupportsDivInPlace<G, G2>::value &&
                        internal::SupportsMulInPlace<G, G2>::value>* = nullptr>
-  constexpr auto& operator/=(const G2& other) {
+  constexpr G& operator/=(const G2& other) {
     G* g = static_cast<G*>(this);
     return g->MulInPlace(other.Inverse());
   }
@@ -100,7 +100,7 @@ class AdditiveGroup : public AdditiveSemigroup<G> {
   template <
       typename G2,
       std::enable_if_t<internal::SupportsSubInPlace<G, G2>::value>* = nullptr>
-  constexpr auto& operator-=(const G2& other) {
+  constexpr G& operator-=(const G2& other) {
     G* g = static_cast<G*>(this);
     return g->SubInPlace(other);
   }
@@ -110,7 +110,7 @@ class AdditiveGroup : public AdditiveSemigroup<G> {
       typename G2,
       std::enable_if_t<!internal::SupportsSubInPlace<G, G2>::value &&
                        internal::SupportsAddInPlace<G, G2>::value>* = nullptr>
-  constexpr auto& operator-=(const G2& other) {
+  constexpr G& operator-=(const G2& other) {
     G* g = static_cast<G*>(this);
     return g->AddInPlace(other.Negative());
   }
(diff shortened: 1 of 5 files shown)


TomTaehoonKim:
There is a typo in the first commit message! Fix assignemnt -> assignment please!

tachyon/math/base/groups.h
```diff
@@ -48,7 +48,7 @@ class MultiplicativeGroup : public MultiplicativeSemigroup<G> {
   template <
       typename G2,
       std::enable_if_t<internal::SupportsDivInPlace<G, G2>::value>* = nullptr>
-  constexpr auto& operator/=(const G2& other) {
```
TomTaehoonKim, line 51: What other types could come if we haven't forced to return `G&`?

tachyon/math/base/groups.h
```diff
@@ -96,23 +97,23 @@ class AdditiveGroup : public AdditiveSemigroup<G> {
...
-  constexpr G& operator-=(const G2& other) {
-    G* g = static_cast<G*>(this);
-    return g->SubInPlace(other);
-  }
-
-  // Subtraction in place: a += (-b)
-  template <
-      typename G2,
-      std::enable_if_t<!internal::SupportsSubInPlace<G, G2>::value &&
+      std::enable_if_t<internal::SupportsSubInPlace<G, G2>::value ||
```
TomTaehoonKim, line 111: Can you fix the commit message? I think you have united `operator-=()` not `operator*=()`.

tachyon/math/base/groups_unittest.cc
```diff
@@ -57,6 +57,39 @@ TEST(GroupsTest, DivOverMul) {
...
+    explicit IntInverse(int denominator) : denominator_(denominator) {}
+    IntInverse(const IntInverse& other) : denominator_(other.denominator_) {}
+
+    bool operator==(const IntInverse& other) const {
+      return denominator_ == other.denominator_;
+    }
+
+   private:
+    int denominator_;
+  };
```
TomTaehoonKim, line 73:
```suggestion
  class IntInverse {
   public:
    IntInverse() = default;
    explicit IntInverse(int denominator) : denominator_(denominator) {}
    IntInverse(const IntInverse& other) : denominator_(other.denominator_) {}

    bool operator==(const IntInverse& other) const {
      return denominator_ == other.denominator_;
    }

   private:
    int denominator_ = 0;
  };
```
How about like this as we used to do?

tachyon/math/base/groups_unittest.cc
```diff
@@ -57,6 +57,39 @@ TEST(GroupsTest, DivOverMul) {
...
+    explicit Int(int value) : value_(value) {}
+    Int(const Int& other) : value_(other.value_) {}
+
+    IntInverse Inverse() const { return IntInverse(value_); }
+
+    bool operator==(const Int& other) const { return value_ == other.value_; }
+
+   private:
+    int value_;
+  };
```
TomTaehoonKim, line 87:
```suggestion
  class Int : public MultiplicativeGroup<Int> {
   public:
    Int() = default;
    explicit Int(int value) : value_(value) {}
    Int(const Int& other) : value_(other.value_) {}

    IntInverse Inverse() const { return IntInverse(value_); }

    bool operator==(const Int& other) const { return value_ == other.value_; }

   private:
    int value_ = 0;
  };
```

tachyon/math/base/groups_unittest.cc
```diff
@@ -109,4 +142,37 @@ TEST(GroupsTest, SubOverAdd) {
...
+    explicit Int(int value) : value_(value) {}
+    Int(const Int& other) : value_(other.value_) {}
+
+    IntNegative Negative() const { return IntNegative(-value_); }
+
+    bool operator==(const Int& other) const { return value_ == other.value_; }
+
+   private:
+    int value_;
+  };
```
TomTaehoonKim, line 172:
```suggestion
  class Int : public AdditiveGroup<Int> {
   public:
    Int() = default;
    explicit Int(int value) : value_(value) {}
    Int(const Int& other) : value_(other.value_) {}

    IntNegative Negative() const { return IntNegative(-value_); }

    bool operator==(const Int& other) const { return value_ == other.value_; }

   private:
    int value_ = 0;
  };
```

tachyon/math/base/groups.h
chokobole: Only `G&` can come. 

```c++
T a;
a += b; // a doesn't change its type!
```

TomTaehoonKim (approved):
LGTM

Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #93, opened 2023-10-16
feat: implement rational field

chokobole:
# Description

This PR includes followings:

1. Implement `RationalField`: `RationalField` class can be used to optimize performance by delaying an expensive inverse computation. The motivation is very similar to why you use `JacobianPoint` or others instead of `AffinePoint`.
2. Refactor `Field<F>::BatchInverse()`.(use pointer for out parameter)
3. Add an error message when `Field<F>::BatchInverse()` returning false.
4. Let gtest uses `ToString()` if `operator<<()` is not implemented on type.


tachyon/base/strings/string_util.h
```diff
@@ -114,7 +115,8 @@ std::string VectorToString(const std::vector<T>& data) {
   std::stringstream ss;
   ss << "[";
   for (size_t i = 0; i < data.size(); ++i) {
-    ss << data[i];
+    // NOTE(chokobole): This is a trick to call `operator<<()` or `ToString()`.
```
Insun35, line 118:
```suggestion
    // NOTE(chokobole): This is a trick to call |operator<<()| or |ToString()|.
```

Insun35 (commented):
Else, looks good to me!

TomTaehoonKim (approved):
LGTM

PR #94, opened 2023-10-16
refac(math): refactor univariate polynomial parts

chokobole:
# Description

- change return type for `EvaluateAllLagrangeCoefficients()` from `std::vector<F>` to `DenseCoefficients<F, MaxDegree>`
- implement `operator[]` instead of `Get()` for `UnivariateXXXCoefficients`.


Insun35 (approved):
LGTM

PR #95, opened 2023-10-17
feat: implement range

chokobole:
# Description

This PR implements `Range` class that covers all these intervals.

- inclusive and inclusive (a ≤ x ≤ b)
- inclusive and exlusive (a ≤ x < b)
- exclusive and inclusive (a < x ≤ b)
- exclusive and exlusive (a < x < b)

In addition, `UniformXXX()` function now makes use of this `Range`.


TomTaehoonKim (approved):
LGTM!

PR #90, opened 2023-10-14
feat: implement square root algorithms

chokobole:
# Description

This implements `Shanks` and `Tonelli-Shanks` algorithms for prime field square root.
They are also implemented in arkworks and we can add more algorithms listed in [here](https://eprint.iacr.org/2012/685.pdf) later on.

PR #119, opened 2023-10-25
feat(math): create point from x coordinate if (x, y) belongs to a curve point

chokobole:
# Description

This PR implements point creation from x coordinate using #90.
This also adds missing `IsOnCurve` tests to `JacobianPoint`, `ProjectivePoint` and `PointXYZZ`.

### DIFF
--- a/tachyon/math/elliptic_curves/short_weierstrass/affine_point.h
+++ b/tachyon/math/elliptic_curves/short_weierstrass/affine_point.h
@@ -1,6 +1,7 @@
 #ifndef TACHYON_MATH_ELLIPTIC_CURVES_SHORT_WEIERSTRASS_AFFINE_POINT_H_
 #define TACHYON_MATH_ELLIPTIC_CURVES_SHORT_WEIERSTRASS_AFFINE_POINT_H_
 
+#include <optional>
 #include <string>
 #include <type_traits>
 #include <utility>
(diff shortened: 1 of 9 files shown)


tachyon/math/elliptic_curves/short_weierstrass/affine_point.h
```diff
@@ -61,6 +62,13 @@ class AffinePoint<_Curve, std::enable_if_t<_Curve::kIsSWCurve>> final
     return ret;
   }
 
+  constexpr static std::optional<AffinePoint> CreateFromX(const BaseField& x,
+                                                          bool pick_bigger) {
+    AffinePoint point;
+    if (!Curve::GetPointFromX(x, pick_bigger, &point)) return std::nullopt;
```
dongchangYoo, line 68: Typically, when selecting a `Point`, one cannot determine whether their y-value is large or small (requiring additional computation), so in Ethereum signatures, the y-value is chosen based on "odd/even." Don't you think we need to address this?

tachyon/math/elliptic_curves/short_weierstrass/sw_curve.h
```diff
@@ -37,6 +40,45 @@ class SWCurve {
     Config::Init();
   }
 
+  // Attempts to construct an affine point given an |x| coordinate. The
+  // point is not guaranteed to be in the prime order subgroup.
+  // If |pick_bigger| is set to true, it chooses a bigger y coordinate.
+  template <typename PointTy>
+  constexpr static bool GetPointFromX(const BaseField& x, bool pick_bigger,
```
dongchangYoo, line 47: Typically, when selecting a Point, one cannot determine whether their y-value is large or small (requiring additional computation), so in Ethereum signatures, the y-value is chosen based on "odd/even." Don't you think we need to address this?

chokobole: I took it from arkworks, but I think your suggestion is nicer!

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #122, opened 2023-10-27
perf: pass parameter only types by value

chokobole:
# Description

This PR passes "parameter-only-types" by value not by reference!

See https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/


TomTaehoonKim (approved):
LGTM

PR #126, opened 2023-10-29
refac: rename batch to multi

chokobole:
# Description

- Unlike `BatchInverse`, `BatchScalarMul` doesn't have a benifit itself when doing in batch. It just parallelize if openmp feature is turned on. So to deliver its meaning clearer, this commit renames the functions.

- This PR also generalizes `ScalarMul` and `MultiScalarMul` by accepting more type as function argumetns.


Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #138, opened 2023-10-31
feat!: remove MaxDegree and use MaxSize only

chokobole:
# Description

Previously, for example, when creating coefficients, you need to pass arguments in terms of degree.

```c++
KZGCommitmentScheme kzg;
DensePoly::Random(kzg.n() - 1); // This accepts degree not a size.
```

To reduce such a confusion, this commit use size throughout the codes.

### DIFF
--- a/tachyon/math/polynomials/univariate/univariate_evaluations.h
+++ b/tachyon/math/polynomials/univariate/univariate_evaluations.h
@@ -23,7 +23,7 @@ namespace tachyon {
 namespace math {
 namespace internal {
 
-template <typename F, size_t MaxDegree>
+template <typename F, size_t N>
 class UnivariateEvaluationsOp;
 
 }  // namespace internal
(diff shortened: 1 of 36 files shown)


tachyon/math/polynomials/univariate/univariate_evaluations.h
```diff
@@ -36,49 +36,49 @@ class UnivariateEvaluationsOp;
...
-  constexpr static size_t kMaxDegree = MaxDegree;
+  constexpr static size_t kMaxSize = N;
 
   using Field = F;
 
   constexpr UnivariateEvaluations() : UnivariateEvaluations({F::Zero()}) {}
   constexpr explicit UnivariateEvaluations(const std::vector<F>& evaluations)
       : evaluations_(evaluations) {
-    CHECK_LE(Degree(), MaxDegree);
+    CHECK_LE(evaluations_.size(), kMaxSize);
```
dongchangYoo, line 50: `UnivariateEvaluations` does not add or remove leading zeros, so it must receive `evaluations` of the exact length. And if `kMaxSize` represents 2^s, then the formula is correct, but if it means 2^n, we should change it to `CHECK_EQ`. 

In this PR, we interpret `kMaxSize` as 2^s, and I will correct it in the PR that assigns Size to PCS.

tachyon/math/polynomials/univariate/univariate_evaluations.h
```diff
@@ -36,49 +36,49 @@ class UnivariateEvaluationsOp;
...
   }
   constexpr explicit UnivariateEvaluations(std::vector<F>&& evaluations)
       : evaluations_(std::move(evaluations)) {
-    CHECK_LE(Degree(), MaxDegree);
+    CHECK_LE(evaluations_.size(), kMaxSize);
```
dongchangYoo, line 54: UnivariateEvaluations does not add or remove leading zeros, so it must receive evaluations of the exact length. And if kMaxSize represents 2^s, then the formula is correct, but if it means 2^n, we should change it to CHECK_EQ.

In this PR, we interpret kMaxSize as 2^s, and I will correct it in the PR that assigns Size to PCS.

tachyon/math/polynomials/univariate/univariate_evaluations.h
dongchangYoo: You can resolve this by yourself!

chokobole:
Close this issue because If i think it intuitively the polynomial should be defined with a degree and If we mix up the polynomial with a size, there will be conflict between them. Even now it's happening when thinking about multivariate polynomial.

```c++
// If we define this polynomial with size `N`, then the constraint would be d - 1 < N
// x^{d-1} - 1 

// Then what about the multivariate case? a + b < ??
// x^a * y^b - 1 
```

PR #156, opened 2023-11-06
ci: add node binding and py binding test

Insun35:
# Description

Add Node binding test and Python binding test for both Linux and MacOS.

PR #157, opened 2023-11-06
fix: fix broken binding build

chokobole:
# Description

This PR fixes CI since [this](https://github.com/kroma-network/tachyon/pull/156). The symptom is it can deduce `Negative()` signature. The reason is still unclear to me. I believe this is a bug or lack of feature in `NodeCppConstructor`. This PR doesn't solve the issue directly but rather solve in another way. It binds `operator-()` unary operator directly and removes `Negative()` method, which also might makes user confused since `Negative()` is not supported, even when `operator-()` is supported!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,12 +48,12 @@ jobs:
       - name: Test Node Binding
         run: |
           cd tachyon/node/test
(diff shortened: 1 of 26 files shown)


dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

PR #167, opened 2023-11-13
refac(zk): unify query method names

Insun35:
# Description

Unify the method names so that each `ProverQuery` and `VerifierQuery` can be used in common functions like `ConstructIntermediateSets` in a multi-point opening scheme on the prover and verifier sides.

`ConstructIntermediateSets` function is a function commonly used by Prover and Verifier.
IntermediateSets are created independently from each other with its own Query.


dongchangYoo (approved):
LGTM

chokobole (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #172, opened 2023-11-27
refac(zk): rename for better readability

chokobole:
# Description

As we talked internally, for better readability, we decided to rename few class and variable names.

### Column

AS-IS: This is used for 2 purposes.
- To locate the column, so existing `Column` contains `index`, `type`, and `phase`.
- To retrieve the evaluation vector.

TO-BE: Existing `Column` is renamed to `ColumnKey`. Now `Column` is used to retrieve the evaluation vector.

### LookupTable

AS-IS: Under permutation directory, there's a `LookupTable`. The authors intended to name it like this because it actually tries to look up the original delta omega when committing permutation argument. But this name sounds not good because it's under permutation directory! 

TO-BE: `LookupTable` is renamed to `UnpermutedTable`

### Rows

AS-IS: In `UnpermutedTable`, The authors first think that the table consists of the set of rows. But when you see it broadly, it should be interpreted as the set of columns. The table consists of `advice`, `instance` and `fixed` columns in terms of Halo2.

TO-BE: `Rows` is renamed to `Cols`.

### Table

AS-IS: We want to consider Plonkish as filling in table according to constraints in the frontend and producing a proof based on this table according to a method established by a proving scheme. Then table should point to the set of columns not just for lookup table. 

TO-BE: As it is written in comment, an existing `Table` is renamed to be `LookupTable`.


tachyon/zk/plonk/circuit/column_key.h
```diff
@@ -20,18 +20,20 @@
...
-  ColumnData() = default;
-  ColumnData(ColumnType type, size_t index) : type_(type), index_(index) {}
+  ColumnKeyData() = default;
+  ColumnKeyData(ColumnType type, size_t index) : type_(type), index_(index) {}
 
   ColumnType type() const { return type_; }
   size_t index() const { return index_; }
 
-  bool operator==(const ColumnData& other) const {
+  bool operator==(const ColumnKeyData& other) const {
```
dongchangYoo, line 31: How about `ColumnKeyBase' instead of 'ColumnKeyData'?

tachyon/zk/plonk/permutation/unpermuted_table.h
```diff
@@ -29,8 +29,8 @@ class UnpermutedTable {
  public:
   using F = typename PCSTy::Field;
   using Domain = typename PCSTy::Domain;
-  using Rows = std::vector<F>;
-  using Table = std::vector<Rows>;
+  using Cols = std::vector<F>;
```
dongchangYoo, line 32: I think `Col` or `Column` is more appropriate than 'Cols' for `std::vector<F>`, as it represents a single column in a table.

TomTaehoonKim (approved):
LGTM

Insun35 (approved):
LGTM

PR #181, opened 2023-11-30
refac(zk): create verifier

chokobole:
# Description

This PR creates `Verifier` class which is opposite to `Prover`. This class will be used when implementing verification and some of codes that can be called from a verifier side, are changed within this PR.

### DIFF
--- a/tachyon/zk/base/entity.h
+++ b/tachyon/zk/base/entity.h
@@ -1,0 +1,58 @@
+// Copyright 2020-2022 The Electric Coin Company
+// Copyright 2022 The Halo2 developers
+// Use of this source code is governed by a MIT/Apache-2.0 style license that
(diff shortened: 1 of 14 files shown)


tachyon/zk/base/entity.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef TACHYON_ZK_BASE_ENTITY_H_
+#define TACHYON_ZK_BASE_ENTITY_H_
+
+#include <memory>
+#include <utility>
+
+#include "tachyon/zk/transcript/transcript.h"
+
+namespace tachyon::zk {
+
```
Insun35, line 16: I think it might be helpful to add some simple comments that the `Entity` is the parent class that encompasses both prover and verifier.

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #190, opened 2023-12-08
fix(zk): fix bug on constraint system

chokobole:
# Description

This PR fixes bug of `ConstraintSystem` while adding a unittest for it.

### DIFF
--- a/tachyon/zk/plonk/circuit/phase.h
+++ b/tachyon/zk/plonk/circuit/phase.h
@@ -36,6 +36,10 @@ class TACHYON_EXPORT Phase {
 
   bool operator==(Phase other) const { return value_ == other.value_; }
   bool operator!=(Phase other) const { return value_ != other.value_; }
+  bool operator>(Phase other) const { return value_ > other.value_; }
+  bool operator>=(Phase other) const { return value_ >= other.value_; }
+  bool operator<(Phase other) const { return value_ < other.value_; }
+  bool operator<=(Phase other) const { return value_ <= other.value_; }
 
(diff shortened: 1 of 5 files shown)


tachyon/zk/plonk/circuit/phase.h
```diff
@@ -34,8 +34,8 @@ class TACHYON_EXPORT Phase {
 
   uint8_t value() const { return value_; }
 
-  bool operator==(const Phase& other) const { return value_ == other.value_; }
-  bool operator!=(const Phase& other) const { return value_ != other.value_; }
+  bool operator==(Phase other) const { return value_ == other.value_; }
```
TomTaehoonKim, line 37: Can you add a note that passing value here is on purpose?

chokobole: It's already specified at the comment above the class! Otherwise, you should put these comments everywhere!

tachyon/zk/plonk/circuit/phase.h
```diff
@@ -47,7 +47,7 @@ constexpr static Phase kFirstPhase = Phase(0);
 constexpr static Phase kSecondPhase = Phase(1);
 
 template <typename H>
-H AbslHashValue(H h, const Phase& phase) {
+H AbslHashValue(H h, Phase phase) {
```
TomTaehoonKim, line 54: Can you add a note that passing value here is on purpose?

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #199, opened 2023-12-13
refac: create zero or random poly from domain

chokobole:
# Description

Sometimes, we had a mistake while creating `Evals::Random()` or `Evals::UnsafeZero()` by filing it with a size not a degree. To reduce such mistakes, we shouldn't depend on it. Rather, we should create such polynomial through domain because domain knows degree required.


dongchangYoo (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #203, opened 2023-12-15
feat: support halo2 proof encoding

chokobole:
# Description

Halo2 encodes point with a compression making use of a spare bit of base field. But our copyable was a just memory copy including y and infinity flag. To match with an outcome of Halo2 and furthermore support various encoding scheme, `Transcript` class delegates encoding to a child class


TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

Insun35 (commented):
Plz check the CI failure. Else, Looks good to me.

PR #207, opened 2023-12-19
refac: change return type of `EvaluateAllLagrangeCoefficients`

chokobole:
# Description

If we return it as a `DenseCoeffs`, we have to check whether leading coefficients are zero or not, even though we expect it to have zeros. Therefore, we change its return type to `std::vector<F>` to ensure this. FYI, Original arkworks codes use `Vec<F>` as a return type.

See
https://github.com/arkworks-rs/algebra/blob/d9527c832888ffbd9231999b30a700ce3e572c16/poly/src/domain/mod.rs#L156.


Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #209, opened 2023-12-20
refac(zk): replace `RotateOmega()` with `GetElement()`

chokobole:
# Description

TODO: I think we don't need even multplying x if we can create domain using `Coset()` once we get the x.

### DIFF
--- a/tachyon/zk/plonk/circuit/rotation.h
+++ b/tachyon/zk/plonk/circuit/rotation.h
@@ -50,13 +50,9 @@ class TACHYON_EXPORT Rotation {
     return size_t{base::bit_cast<uint32_t>(value % size)};
   }
 
-  template <typename Domain, typename F = typename Domain::F>
+  template <typename Domain, typename F>
   F RotateOmega(const Domain* domain, const F& point) const {
-    if (value_ >= 0) {
-      return point * domain->group_gen().Pow(value_);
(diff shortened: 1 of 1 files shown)


Insun35 (approved):
LGTM

tachyon/zk/plonk/circuit/rotation.h
```diff
@@ -50,15 +50,6 @@ class TACHYON_EXPORT Rotation {
...
   }
 
-  template <typename Domain, typename F = typename Domain::F>
-  F RotateOmega(const Domain* domain, const F& point) const {
-    if (value_ >= 0) {
-      return point * domain->group_gen().Pow(value_);
-    } else {
-      return point * domain->group_gen_inv().Pow(-1 * value_);
-    }
-  }
```
TomTaehoonKim, line 60:
```suggestion
  template <typename Domain, typename F = typename Domain::F>
  F RotateOmega(const Domain* domain, const F& point) const {
    return point * domain->GetElement(value_);
  }
```
I think it's better to keep `RotateOmega` as `Rotation`. How about just fixing the code? (GetElement already handles with the signs.)

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #216, opened 2024-01-06
refac: remove ty suffix

chokobole:
# Description

Initially our team set the policy to put suffix `Ty` to differentiate a template type from real existing type, such as `PrimeFieldTy` vs `PrimeField`. But somehow we have mixed using with and without `Ty` and we determined to not using `Ty` suffix policy any more. This PR doesn't rename all the cases because some are required to add suffix `Ty`.


TomTaehoonKim (approved):
LGTM, but what are some cases that require to add suffix Ty?

chokobole:
> LGTM, but what are some cases that require to add suffix Ty?

Kind of this?

```c++
template <typename Config>
struct SWCurveTraits {
  using BaseField = typename Config::BaseField;
  using ScalarField = typename Config::ScalarField;
  using AffinePointTy = AffinePoint<SWCurve<Config>>;
  using ProjectivePointTy = ProjectivePoint<SWCurve<Config>>;
  using JacobianPointTy = JacobianPoint<SWCurve<Config>>;
  using PointXYZZTy = PointXYZZ<SWCurve<Config>>;
};
```

dongchangYoo:
It would be good to address the modifications for the following three types in this PR as well.
- ColumKeyTy
- QueryDataTy
- CIrcuitTy

chokobole:
> It would be good to address the modifications for the following three types in this PR as well.
> 
> * ColumKeyTy
> * QueryDataTy
> * CIrcuitTy

I'll cover more cases. Thanks!

dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

PR #220, opened 2024-01-07
refac: implement linear combination

chokobole:
# Description

This replaces `Polynomial`'s `Linearize(InPlace)()` methods and manual linear combinations.
    
This method covers all 3 cases.
    
- AdditiveSemigroup's scalar multiplication e.g, affine point
- Ring's multiplication and addition e.g, polynomial
- Field's multiplication and addition e.g, prime field

### DIFF
--- a/tachyon/math/base/semigroups.h
+++ b/tachyon/math/base/semigroups.h
@@ -1,6 +1,9 @@
 #ifndef TACHYON_MATH_BASE_SEMIGROUPS_H_
 #define TACHYON_MATH_BASE_SEMIGROUPS_H_
 
+#include <stddef.h>
+#include <stdint.h>
+
 #include <algorithm>
 #include <vector>
 
@@ -9,6 +12,7 @@
 #include "tachyon/base/bits.h"
 #include "tachyon/base/containers/adapters.h"
 #include "tachyon/base/containers/container_util.h"
+#include "tachyon/base/logging.h"
 #include "tachyon/base/parallelize.h"
 #include "tachyon/base/types/always_false.h"
 #include "tachyon/math/base/big_int.h"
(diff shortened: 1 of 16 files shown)


dongchangYoo:
In general, "Linear Combination" means inner product of two vectors as following link; https://en.wikipedia.org/wiki/Linear_combination.

Insun35 (approved):
LGTM

dongchangYoo:
In all citation papers of the Plonk paper, the term `random linear combination` was used as a common noun, referring to the combination with a random vector as a `random linear combination`.

There are two recommandations.
- use 'LinearCombinationWithRandomPowers" instead of "LinearCombination" (Is it too long name?)
- overloading functions
  - `F LinearCombination(const Container1<T1>& vec1, const Contatiner<T2>& vec2)` for calculate inner product
  - `F LinearCombination(const Contatiner1<T>& vec, const Point& r)` -> F for calculate `Sum(vec[i] * r^i)`

tachyon/math/base/semigroups.h
```diff
@@ -380,6 +381,76 @@ class AdditiveSemigroup {
...
 
+  // Linear combination
+  // - forward: a₀ * rⁿ⁻¹ + a₁ * rⁿ⁻² + ... + aₙ₋₁
+  // - backward: a₀ + a₁ * r + ... + aₙ₋₁ * rⁿ⁻¹
+  // NOTE(chokobole): For a performance reason, I would recommend the
+  // |LinearCombinationInPlace()| if possible.
+  template <bool Forward, typename Container, typename T,
+            typename ReturnTy =
+                typename internal::AdditiveSemigroupTraits<G>::ReturnTy>
+  constexpr static ReturnTy LinearCombination(const Container& values,
```
TomTaehoonKim, line 392: Unlike LinearCombinationInPlace, did you intend to return Zero when size of values is 0?

TomTaehoonKim (changes requested):

chokobole:
> In all citation papers of the Plonk paper, the term `random linear combination` was used as a common noun, referring to the combination with a random vector as a `random linear combination`.
> 
> There are two recommandations.
> 
> * use 'LinearCombinationWithRandomPowers" instead of "LinearCombination" (Is it too long name?)
> * overloading functions
>   
>   * `F LinearCombination(const Container1<T1>& vec1, const Contatiner<T2>& vec2)` for calculate inner product
>   * `F LinearCombination(const Contatiner1<T>& vec, const Point& r)` -> F for calculate `Sum(vec[i] * r^i)`

Can't we just think that `RandomLinearCombination` without `Random`? I have to think more about alternative suggestion because we aren't working on linear algebra, and we refer to it with `InnerProduct` instead mostly when considering `IPA`?

tachyon/math/base/semigroups.h
chokobole: Yes, I believe `LinearCombination()` should do like that and `LinearCombinationInPlace()` can't return anything if empty arrays are given.

dongchangYoo:
> > In all citation papers of the Plonk paper, the term `random linear combination` was used as a common noun, referring to the combination with a random vector as a `random linear combination`.
> > There are two recommandations.
> > 
> > * use 'LinearCombinationWithRandomPowers" instead of "LinearCombination" (Is it too long name?)
> > * overloading functions
> >   
> >   * `F LinearCombination(const Container1<T1>& vec1, const Contatiner<T2>& vec2)` for calculate inner product
> >   * `F LinearCombination(const Contatiner1<T>& vec, const Point& r)` -> F for calculate `Sum(vec[i] * r^i)`
> 
> For the first recommendation, it seems long to me as well. For the second recommendation, I have to think more about alternative suggestions because we aren't working on linear algebra, and we refer to it with `InnerProduct` instead mostly when considering `IPA`?
> 
> Can't we just think that `RandomLinearCombination` without `Random`? For me, it doesn't sound that strange!

This probably means that they you keep the current function name as it is, right?
Understood. This conversation thread has been resolved.

dongchangYoo (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #221, opened 2024-01-07
feat(math): improve get successive methods

chokobole:
# Description

This PR implements 2 methods.

- `GetSuccessivePowers()`
- `GetSuccessiveScalarMuls()`

Additionally, `GetSuccessivePowers()` gets improved by accepting 1 more parameter.

### DIFF
--- a/tachyon/math/base/semigroups.h
+++ b/tachyon/math/base/semigroups.h
@@ -170,7 +170,8 @@ class MultiplicativeSemigroup {
       size_t size, const G& generator, const G& c = G::One()) {
     std::vector<ReturnTy> ret;
     ret.resize(size);
-    size_t num_elems_per_thread = base::GetNumElementsPerThread(ret);
+    size_t num_elems_per_thread =
+        base::GetNumElementsPerThread(ret, kDefaultParallelThreshold);
     OPENMP_PARALLEL_FOR(size_t i = 0; i < size; i += num_elems_per_thread) {
       ReturnTy pow = c * generator.Pow(i);
       for (size_t j = i; j < i + num_elems_per_thread && j < size; ++j) {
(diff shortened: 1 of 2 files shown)


Insun35 (approved):
LGTM

dongchangYoo:
The `DistributePowersAndMulByConst` is similar to `GetSuccessivePowers`. 
Could you fix `DistributePowersAndMulByConst` to use `GetSuccessivePowers`? (If you don't think this change doesn't match this PR)

chokobole:
> The `DistributePowersAndMulByConst` is similar to `GetSuccessivePowers`. � Could you fix `DistributePowersAndMulByConst` to use `GetSuccessivePowers`? (If you don't think this change doesn't match this PR)

I wanted to but `DistributePowersAndMulByConst` takes `Polynomial` or `Evaluations` and when indexing it, it returns a pointer not a reference. That's why i remain it untouched.

tachyon/math/base/semigroups.h
```diff
@@ -301,6 +301,23 @@ class AdditiveSemigroup {
     }
   }
 
+  // generator: g
+  // return: [c, c * g, c * 2 * g, ..., c * (|size| - 1) * g]
+  constexpr static std::vector<ReturnTy> GetSuccessiveScalarMuls(
+      size_t size, const G& generator, const G& c = G::One()) {
```
TomTaehoonKim, line 310: Does c here also have something to do with the coset of the evaluataion domain? I'm not sure if we need `c` for the `AdditiveSemigroup` case

TomTaehoonKim (changes requested):

tachyon/math/base/semigroups.h
chokobole: That would be coset or tau! and it reminds me that I should apply this when setting up kzg parameters.

chokobole: This is because it is a method of `AdditiveSemigroup` which corresponds to `GetSuccessivePowers()` of `MultiplicativeSemigroup`. As you said, there's no usecase.

TomTaehoonKim (approved):
LGTM

PR #239, opened 2024-01-18
refac(zk): re-define commit methods of `ProverBase`

dongchangYoo:
The original `Commit()` commits the input, then written to the transcript. If this method is called in parallel, it becomes impossible to guarantee the order of the internal state of the transcript and the proof array. In this commit, 3 methods were introduced as follow:
- `Commit()` returns a commitment.
- `CommitAndWriteToTranscript()` writes a commitment to transcript.
- `CommitAndWriteToProof()` writes commit a commitment to transcript and proof.

### DIFF
--- a/tachyon/zk/plonk/circuit/examples/BUILD.bazel
+++ b/tachyon/zk/plonk/circuit/examples/BUILD.bazel
@@ -1,4 +1,4 @@
-load("//bazel:tachyon_cc.bzl", "tachyon_cc_library", "tachyon_cc_unittest")
+load("//bazel:tachyon_cc.bzl", "tachyon_cc_library", "tachyon_cc_test")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -21,7 +21,9 @@ tachyon_cc_library(
     deps = ["//tachyon/zk/plonk/circuit"],
 )
 
-tachyon_cc_unittest(
+# TODO(dongchangYoo): This is failed in CI because of timeout, 60 secs.
+# Change |tachyon_cc_test| to |tachyon_cc_unittest| once CI can run within the timeout.
+tachyon_cc_test(
     name = "examples_unittests",
     srcs = [
         "simple_circuit_unittest.cc",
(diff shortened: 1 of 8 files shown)


chokobole (approved):
LGTM

TomTaehoonKim:
@dongchangYoo In the first commit, could you fix "then written" -> to "then writes"?

tachyon/zk/plonk/circuit/examples/BUILD.bazel
```diff
@@ -21,7 +21,7 @@ tachyon_cc_library(
     deps = ["//tachyon/zk/plonk/circuit"],
 )
 
-tachyon_cc_unittest(
+tachyon_cc_test(
```
TomTaehoonKim, line 24: Can you add a TODO comment for this change please?

TomTaehoonKim (changes requested):

tachyon/zk/plonk/circuit/examples/BUILD.bazel
```diff
@@ -21,7 +21,9 @@ tachyon_cc_library(
     deps = ["//tachyon/zk/plonk/circuit"],
 )
 
-tachyon_cc_unittest(
+# TODO(dongchangYoo): Change |tachyon_cc_test| to
+# |tachyon_cc_unittest| after increasing the timeout for unittests.
```
chokobole, line 25: This comment doesn't make sense. You need to say in this way

```suggestion
# TODO(dongchangYoo): This is failed in CI because of timeout, 60 secs. 
# Change |tachyon_cc_test| to |tachyon_cc_unittest| once CI can run within the timeout.
```

chokobole (changes requested):

tachyon/zk/plonk/circuit/examples/BUILD.bazel
dongchangYoo: I mistakenly thought that the timeout would eventually be extended. 
I have edited the comment.

TomTaehoonKim (approved):
LGTM

chokobole (approved):
LGTM

Insun35 (approved):
LGTM

PR #240, opened 2024-01-19
perf(zk): remove redundant assignment

TomTaehoonKim:
# Description

Since `acc` is contained in a register, addition can be done through 1 instruction. If you do `+=`, it returns a pointer, so on return inside lambda of `std::accumulate`, when assigning to `acc`, more instructions are needed such as loading address and copying values of pointer.


chokobole (approved):
LGTM

chokobole:
I checked in https://godbolt.org/z/4oojG6fM9 but it turns out that the compiled outputs actually are same. Sorry but for me it seems still better!

dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

ashjeong:
LGTM

PR #251, opened 2024-01-25
fix(zk): match the stringify method for a non-zero `Phase` with the rust-halo2

dongchangYoo:
This PR solved the issue where the rule for converting a non-zero `ColumnKey` to a string was different from rust-halo2.
The `RustDebugStringifier` does not export `Phase` of advice column while the rust-halo2 does it when the phase is non-zero. (The `ColumnType` in rust-halo2 includes `Phase`.)

[FYI] Example of stringify rules of the rust halo2:
- `ColumnType` with `Phase(0)` => "Advice",
- `ColumnType` with the other phases => "Advice { phase: Phase(1) }"

This affects the representative value of the verifying key.

### DIFF
--- a/tachyon/zk/plonk/circuit/column_key_stringifier.h
+++ b/tachyon/zk/plonk/circuit/column_key_stringifier.h
@@ -15,6 +15,22 @@
 
 namespace tachyon::base::internal {
 
+struct AdviceColumnType {
+  zk::Phase phase;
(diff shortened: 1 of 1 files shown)


chokobole (approved):
LGTM

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #252, opened 2024-01-26
perf: allocate poly in c++ memory when proving

chokobole:
# Description

This PR fixes to allocate memory for `Polynomial` and `Evaluations` inside c++ memory so that it can avoid copying rust memory which usually takes a lot.

### DIFF
--- a/vendors/halo2/src/bn254_shplonk_prover.cc
+++ b/vendors/halo2/src/bn254_shplonk_prover.cc
@@ -5,7 +5,11 @@
 #include "tachyon/math/polynomials/univariate/univariate_evaluation_domain_factory.h"
 #include "tachyon/rs/base/container_util.h"
 #include "tachyon/rs/base/rust_vec_copyable.h"
+#include "vendors/halo2/include/bn254_evals.h"
+#include "vendors/halo2/include/bn254_poly.h"
 #include "vendors/halo2/src/bn254.rs.h"
+#include "vendors/halo2/src/bn254_evals_impl.h"
+#include "vendors/halo2/src/bn254_poly_impl.h"
 #include "vendors/halo2/src/bn254_shplonk_prover_impl.h"
 #include "vendors/halo2/src/bn254_shplonk_proving_key_impl.h"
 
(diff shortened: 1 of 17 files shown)


vendors/halo2/src/bn254_shplonk_prover.cc
```diff
@@ -27,20 +31,34 @@ rust::Box<G1JacobianPoint> DoCommit(
       reinterpret_cast<G1JacobianPoint*>(result));
 }
 
-std::vector<math::bn254::Fr> ReadFrs(uint8_t* vec_ptr, size_t i) {
+std::vector<math::bn254::Fr> ReadFrsToRemove(uint8_t* vec_ptr, size_t i) {
```
TomTaehoonKim: Do we need this temporary change?

vendors/halo2/src/bn254_shplonk_prover.cc
```diff
@@ -27,20 +31,34 @@ rust::Box<G1JacobianPoint> DoCommit(
...
   base::Buffer buffer(&vec_ptr[num_bytes * i], num_bytes);
   rs::RustVec vec;
   CHECK(buffer.Read(&vec));
   return vec.ToVec<math::bn254::Fr>();
 }
 
+PCS::Evals ReadEvalsToRemove(uint8_t* vec_ptr, size_t i) {
+  return PCS::Evals(ReadFrsToRemove(vec_ptr, i));
+}
+
```
TomTaehoonKim: Do we need this temporary change?

TomTaehoonKim (changes requested):

vendors/halo2/src/bn254_shplonk_prover.cc
chokobole: This is because there's symbol conflict `ReadEvals`!

vendors/halo2/src/bn254_shplonk_prover.cc
chokobole: This is coined from below reason. `ReadEvals` anyway should be renamed and `ReadFrs` are renamed correspondently.

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

PR #266, opened 2024-01-30
fix(zk): clone config before move

TomTaehoonKim:
# Description

I found a bug while implementing fibonacci3 circuit, where `config` is being moved twice, but `Clone` wasn't called on the first case. This PR fixes such bug.

### DIFF
--- a/tachyon/zk/plonk/layout/floor_planner/v1/v1_floor_planner.h
+++ b/tachyon/zk/plonk/layout/floor_planner/v1/v1_floor_planner.h
@@ -47,7 +47,7 @@ class V1FloorPlanner : public FloorPlanner<Circuit> {
     MeasurementPass<F> measure;
     {
       V1Pass<F> pass(&measure);
-      circuit.WithoutWitness()->Synthesize(std::move(config), &pass);
+      circuit.WithoutWitness()->Synthesize(config.Clone(), &pass);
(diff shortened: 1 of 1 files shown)


tachyon/zk/plonk/layout/floor_planner/v1/v1_floor_planner.h
```diff
@@ -47,7 +47,7 @@ class V1FloorPlanner : public FloorPlanner<Circuit> {
     MeasurementPass<F> measure;
     {
       V1Pass<F> pass(&measure);
-      circuit.WithoutWitness()->Synthesize(std::move(config), &pass);
+      circuit.WithoutWitness()->Synthesize(std::move(config.Clone()), &pass);
```
chokobole, line 50: Since `config.Clone()` is prvalue in https://en.cppreference.com/w/cpp/language/value_category, you don't need to do `std::move()`. 

```suggestion
      circuit.WithoutWitness()->Synthesize(config.Clone(), &pass);
```

chokobole (approved):
LGTM

ashjeong (approved):
LGTM

Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #268, opened 2024-01-31
refac: add plonk namespace

Insun35:
Add `zk::plonk` namespace and set `zk::plonk::halo2` as a nested namespace.

This PR should be merged after #269


tachyon/zk/expressions/expression_type.cc
```diff
@@ -15,15 +15,15 @@ std::string_view ExpressionTypeToString(ExpressionType type) {
     case ExpressionType::kConstant:
       return "Constant";
     case ExpressionType::kSelector:
-      return "Selector";
+      return "plonk::Selector";
```
chokobole, line 18:
```suggestion
      return "Selector";
```

tachyon/zk/expressions/expression_type.cc
```diff
@@ -15,15 +15,15 @@ std::string_view ExpressionTypeToString(ExpressionType type) {
...
     case ExpressionType::kFixed:
       return "Fixed";
     case ExpressionType::kAdvice:
       return "Advice";
     case ExpressionType::kInstance:
       return "Instance";
     case ExpressionType::kChallenge:
-      return "Challenge";
+      return "plonk::Challenge";
```
chokobole, line 26:
```suggestion
      return "Challenge";
```

tachyon/zk/plonk/keys/verifying_key.h
```diff
@@ -103,7 +103,7 @@ class VerifyingKey : public Key {
 
   template <typename PCS>
   void SetTranscriptRepresentative(const Entity<PCS>* entity) {
-    halo2::PinnedVerifyingKey<F, C> pinned_verifying_key(entity, *this);
+    plonk::halo2::PinnedVerifyingKey<F, C> pinned_verifying_key(entity, *this);
```
chokobole, line 106:
```suggestion
    halo2::PinnedVerifyingKey<F, C> 
```

chokobole (changes requested):

chokobole (approved):
LGTM

tachyon/zk/plonk/permutation/grand_product_argument.h
```diff
@@ -55,7 +57,7 @@ class GrandProductArgument {
   }
 
  private:
-  FRIEND_TEST(LookupArgumentRunnerTest, ComputePermutationProduct);
+  friend class zk::LookupArgumentRunnerTest_ComputePermutationProduct_Test;
```
ashjeong, line 60: What does this new random class do? There are other instances of private `FRIEND_TEST`s that aren't changed like this like in `selector_compressor.h?`

chokobole: This is what `FRIEND_TEST()` does. 

```c++
FRIEND_TEST(zk::LookupArgumentRunnerTest, ComputePermutationProduct);
// equivalent to
// friend class LookupArgumentRunnerTest_ComputePermutationProduct_Test;
```

I don't know the exact reason, but doing only above can't look up the symbol correctly. So we need a forward declaration above, then we end up with a question. If we do this ourselves, do we actually need `FRIEND_TEST()`? This question lead us to just flatten like above ourselves.

ashjeong (approved):
LGTM

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #278, opened 2024-02-05
fix: handle point at infinity case

chokobole:
# Description

When printing a debug string and when creating a point, Halo2 treats the point at infinity specially. This PR handles both cases.


Insun35 (approved):
LGTM

ashjeong (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #283, opened 2024-02-06
refac: add `at` to polynomials

chokobole:
# Description

In order to access an element in a polynomial, you need to use the `[]` operator which in turn gives you a pointer of the field. It's likely we carelessly use the `[]` operator without a boundary check.

An error may show up where the result of the `[]` operator is dereferenced, since if there exists no element at a given index, the `[]` operator will return a nullptr. Here some more concrete examples of error-prone statements:
```c++
const F& value = *poly[idx];
*poly[idx] = F::One();
```

Thus, this commit has made 2 changes to fix this behavior.

1. The return type of the `[]` operator has been changed to a const reference. If there is no element at the given index, a const reference of the zero point of the field is returned.
2. The `at()` method has been added. Similar to STL, the `at()` method does a boundary check. See [at](https://en.cppreference.com/w/cpp/container/vector/at) and [operator[]](https://en.cppreference.com/w/cpp/container/vector/operator_at).

After this commit, we encourage developers to leave a comment noting whether accessing the given index is safe when using `at()`. Comments do not need to be added repetitively to `at()` methods in files that follow an overarching assumption. (like in univariate_evaluation_domain.h)

### DIFF
--- a/tachyon/c/math/polynomials/univariate/bn254_univariate_dense_polynomial_unittest.cc
+++ b/tachyon/c/math/polynomials/univariate/bn254_univariate_dense_polynomial_unittest.cc
@@ -35,16 +35,18 @@ class UnivariateDensePolynomialTest : public testing::Test {
 }  // namespace
 
 TEST_F(UnivariateDensePolynomialTest, Clone) {
-  if (reinterpret_cast<Poly&>(*poly_)[0] != nullptr) {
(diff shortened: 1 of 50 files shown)


TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #284, opened 2024-02-06
fix(math): correct parallel IFFT

chokobole:
# Description

When compacting roots in place, there is a potential for collisions. Our code originally relied on the assumption that the assignments of the new compacted roots ran in order from smallest to largest index, but parallelization faults this assumption.

For example, we expect moves to occur in our roots vector in the order of indexes 0, 1, 2 like so:

0: [0] <- [0]
1: [1] <- [2]
2: [2] <- [4]

However, if the order changes to assign indexes 0, 2, 1 instead due to parallelization, we get this:

0: [0] <- [0]
1: [2] <- [4]
2: [1] <- [2] = [4]

This issue is resolved by compacting roots separately first instead of in place.


TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #286, opened 2024-02-07
fix(zk): add `CHECK` for `LookupAny` as well

TomTaehoonKim:
# Description
While it's not implemented in rust halo2 code, simple selectors are prohibited from appearing in expressions in the lookup argument by `ConstraintSystem`. So, `CHECK` is added to `LookupAny` as well, like `Lookup`.

See https://github.com/kroma-network/tachyon/blob/92f3550bc62dc5d10cfd295c0996b6620f9d46f5/tachyon/zk/expressions/evaluator/selector_replacer.h#L42-L43 and
https://github.com/kroma-network/halo2/blob/7d0a36990452c8e7ebd600de258420781a9b7917/halo2_proofs/src/plonk/circuit.rs#L1563-L1580

### DIFF
--- a/tachyon/zk/plonk/constraint_system/constraint_system.h
+++ b/tachyon/zk/plonk/constraint_system/constraint_system.h
@@ -163,6 +163,12 @@ class ConstraintSystem {
     LookupPairs<std::unique_ptr<Expression<F>>> pairs =
         std::move(callback).Run(cells);
 
+    for (const LookupPair<std::unique_ptr<Expression<F>>>& pair : pairs) {
+      CHECK(!pair.input()->ContainsSimpleSelector())
+          << "expression containing simple selector "
+             "supplied to lookup argument";
+    }
+
     lookups_.emplace_back(name, std::move(pairs));
     return lookups_.size() - 1;
   }
--- a/tachyon/zk/plonk/constraint_system/constraint_system.h
+++ b/tachyon/zk/plonk/constraint_system/constraint_system.h
@@ -23,6 +23,7 @@
 #include "tachyon/base/containers/container_util.h"
 #include "tachyon/base/containers/contains.h"
 #include "tachyon/base/functional/callback.h"
+#include "tachyon/base/logging.h"
 #include "tachyon/base/strings/string_util.h"
 #include "tachyon/zk/base/row_index.h"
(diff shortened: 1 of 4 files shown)


tachyon/zk/plonk/constraint_system/constraint_system.h
```diff
@@ -163,6 +163,12 @@ class ConstraintSystem {
     LookupPairs<std::unique_ptr<Expression<F>>> pairs =
         std::move(callback).Run(cells);
 
+    for (const LookupPair<std::unique_ptr<Expression<F>>>& pair : pairs) {
+      CHECK(!pair.input()->ContainsSimpleSelector())
```
chokobole, line 167: Nice! Next time I want to discuss within the team what kind of statement should be checked in release mode! Since I think it's one of the candidates to be checked in debug mode.

chokobole:
Seeing the https://github.com/kroma-network/tachyon/blob/92f3550bc62dc5d10cfd295c0996b6620f9d46f5/tachyon/zk/expressions/evaluator/selector_replacer.h#L42-L43  it seems already handled. Do you intend to terminate the program early?

tachyon/zk/plonk/constraint_system/constraint_system.h
TomTaehoonKim: I totally agree!

TomTaehoonKim:
@chokobole Yes, my intend was terminating the program early. But now that I'm thinking again, I think maybe it might be better to just remove the `CHECK` from `Lookup` and `LookupAny`, since `CHECK` is already handled in `ReplaceSelectors`.

chokobole:
I don't think this is a `fix` commit. Either one of `refac` or `style` should be chosen. https://github.com/kroma-network/tachyon/pull/286/commits/91e562dd4806e955647bf3449589703f9c266864

TomTaehoonKim:
@chokobole I referred to the first one, but I agree, let's use refac from now on.
Screenshot from 2024-02-08 14-41-57

Insun35 (approved):
LGTM

ashjeong (approved):
LGTM

dongchangYoo (approved):
LGTM

chokobole (approved):
LGTM

PR #289, opened 2024-02-08
fix(zk): read commitments from proof in the correct order

chokobole:
# Description

Unlike other columns, the advice columns have their own phases. The phase iterates from 0, and the advice columns are committed when the current phase is equal to the phases of the advice columns. Therefore, commitments are created in the order of phase. However, commitments should be written to the proof in the order of column. For example, if there are commitments {C₀, C₁, C₂} corresponding to columns 0, 1, 2 and phases {0, 1, 0}, previously, the proof was read in the order of C₀, C₂, C₁, even though the correct order should have been C₀, C₁, C₂. In this commit, the order to read the columns has been corrected.

### DIFF
--- a/tachyon/zk/plonk/halo2/proof_reader.h
+++ b/tachyon/zk/plonk/halo2/proof_reader.h
@@ -53,16 +53,18 @@ class ProofReader {
     const ConstraintSystem<F>& constraint_system =
         verifying_key_.constraint_system();
     proof_.advices_commitments_vec.resize(num_circuits_);
+    size_t num_advice_columns = constraint_system.num_advice_columns();
     for (size_t i = 0; i < num_circuits_; ++i) {
-      proof_.advices_commitments_vec[i].reserve(
-          constraint_system.num_advice_columns());
+      proof_.advices_commitments_vec[i].resize(num_advice_columns);
     }
     proof_.challenges.reserve(constraint_system.challenge_phases().size());
     for (Phase current_phase : constraint_system.GetPhases()) {
       for (size_t i = 0; i < num_circuits_; ++i) {
-        for (Phase phase : constraint_system.advice_column_phases()) {
-          if (current_phase == phase) {
-            proof_.advices_commitments_vec[i].push_back(Read<C>());
+        const std::vector<Phase>& advice_column_phases =
+            constraint_system.advice_column_phases();
+        for (size_t j = 0; j < num_advice_columns; ++j) {
+          if (current_phase == advice_column_phases[j]) {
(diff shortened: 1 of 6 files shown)


tachyon/zk/plonk/halo2/proof_reader.h
```diff
@@ -54,15 +54,17 @@ class ProofReader {
...
-          if (current_phase == phase) {
-            proof_.advices_commitments_vec[i].push_back(Read<C>());
+        const std::vector<Phase>& advice_column_phases =
+            constraint_system.advice_column_phases();
+        for (size_t j = 0; j < advice_column_phases.size(); ++j) {
+          if (current_phase == advice_column_phases[j]) {
+            proof_.advices_commitments_vec[i][j] = Read<C>();
           }
         }
       }
```
TomTaehoonKim, line 70:
```suggestion
    size_t num_advice_columns = constraint_system.num_advice_columns();
    for (size_t i = 0; i < num_circuits_; ++i) {
      proof_.advices_commitments_vec[i].resize(num_advice_columns);
    }
    proof_.challenges.reserve(constraint_system.challenge_phases().size());
    for (Phase current_phase : constraint_system.GetPhases()) {
      for (size_t i = 0; i < num_circuits_; ++i) {
        const std::vector<Phase>& advice_column_phases =
            constraint_system.advice_column_phases();
        for (size_t j = 0; j < num_advice_columns; ++j) {
          if (current_phase == advice_column_phases[j]) {
            proof_.advices_commitments_vec[i][j] = Read<C>();
          }
        }
      }
```
How about like this since we've resized `advice_commitments_vec` with `num_advice_columns`?

chokobole: I don't notice a significant distinction. Could you please clarify the advantage of using `num_advice_columns`?

Insun35 (approved):
LGTM

tachyon/zk/plonk/halo2/proof_reader.h
TomTaehoonKim:
```c++
        for (size_t j = 0; j < num_advice_columns; ++j) {
```
@chokobole It's for this line, where we iterate over `num_advice_columns`. It seems like personal prefernce, but I just thought this is more readable since we resized `advice_commitments_vec` with `num_advice_columns`. You may not apply it since it seems like personal preference.

ashjeong (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #211, opened 2023-12-22
feat: implement SHPlonk VerifyOpeningProof

Insun35:
# Description

Implemented SHPlonk multi-opening proof verifying logic.
See https://eprint.iacr.org/2020/081.pdf

### Refactoring
- Since the `transcripts` have been moved from zk to crypto namespace, 
   it doesn't need to return the value of proofs.
   It can write proofs directly into a transcript.

- Generalized `Poly` to `PolyOracle` for common use in `CreateOpeningProof()` and `VeriryOpeningProof()`.
  `CreateOpeningProof()` uses the `Poly` type, and `VerifyOpeningProof()` uses the `Commitment` type for `PolyOracle`.

- Changed SHPlonk template parameters from Point type to Curve type to use Pairing.

### Fix
- Replaced `absl::flat_hash_map` to `std::vector` since the order of `poly_openings` affects linear combinations.

PR #296, opened 2024-02-14
feat: implement gwc opening scheme

chokobole:
# Description

This PR implements GWC opening scheme which is used by [kroma-prover](https://github.com/kroma-network/kroma-prover/blob/3181f8aa697d2f1d7e35a5d18167280f6bf4a7fc/zkevm/src/prover.rs#L488-L494).


tachyon/crypto/commitments/polynomial_openings_unittest.cc
```diff
@@ -198,39 +198,17 @@ TEST_F(PolynomialOpeningsTest, GroupByPolyOracleAndPoints) {
...
-         grouped_poly_openings_vec[0].poly_openings_vec) {
-      EXPECT_THAT(expected_polys_vec,
-                  testing::Contains(poly_openings.poly_oracle));
-    }
-    for (const PolynomialOpenings<Poly>& poly_openings :
-         grouped_poly_openings_vec[1].poly_openings_vec) {
-      EXPECT_THAT(expected_polys_vec2,
-                  testing::Contains(poly_openings.poly_oracle));
-    }
-  } else {
```
ashjeong, line 217: Does this mean that this else statement never ran previously?

tachyon/crypto/commitments/kzg/BUILD.bazel
```diff
@@ -25,6 +38,22 @@ tachyon_cc_library(
     deps = [":kzg"],
 )
 
+tachyon_cc_library(
+    name = "kzg_family_test",
+    testonly = True,
+    hdrs = ["kzg_family_test.h"],
+    deps = [
+        ":kzg",
```
ashjeong, line 46: If it's a `kzg_family_test`, then why is `kzg_family` not included?

tachyon/crypto/commitments/kzg/kzg_family_test.h
```diff
@@ -0,0 +1,87 @@
...
+#include "tachyon/math/polynomials/univariate/univariate_evaluation_domain_factory.h"
+
+namespace tachyon::crypto {
+
+constexpr static size_t K = 4;
+constexpr static size_t N = size_t{1} << K;
+constexpr static size_t kMaxDegree = N - 1;
+
+template <typename PCS>
+class KZGFamilyTest : public testing::Test {
```
ashjeong, line 26: can we do
```suggestion
class KZGFamilyTest : public FiniteFieldTest<math::bn254::BN254Curve>t {
```
& add `#include "tachyon/math/finite_fields/test/finite_field_test.h"` and delete line 33?

tachyon/zk/base/commitments/gwc_extension.h
```diff
@@ -0,0 +1,217 @@
...
+
+namespace zk {
+
+template <typename Curve, size_t MaxDegree, size_t MaxExtendedDegree,
+          typename Commitment>
+class GWCExtension final
+    : public UnivariatePolynomialCommitmentSchemeExtension<
+          GWCExtension<Curve, MaxDegree, MaxExtendedDegree, Commitment>> {
+ public:
+  // NOTE(dongchangYoo): The following value are pre-determined according to
```
ashjeong, line 40:
```suggestion
  // NOTE(dongchangYoo): The following values are pre-determined according to
```

tachyon/crypto/commitments/polynomial_openings_unittest.cc
chokobole: For the first time we implemented `PolynomialOpenings`, it was flaky. Looking back, we should have thought weird.. but we accepted the changes..

tachyon/crypto/commitments/kzg/BUILD.bazel
chokobole: Yes because when you look at the header, it doesn't include `kzg_family.h`

tachyon/crypto/commitments/kzg/kzg_family_test.h
chokobole: Definitely It would work, but it isn't what I meant. I wanted the type to be only finite field type.

tachyon/crypto/commitments/polynomial_openings_unittest.cc
ashjeong: Ah I see - than maybe changing the title to "fix faulty logic from [#221](https://github.com/kroma-network/tachyon/pull/211)" might be better. I got confused thinking the commit was a new test feature or so.

tachyon/crypto/commitments/kzg/BUILD.bazel
ashjeong: Then isn't `kzg_family_test` misleading as you'd assume from this name that `kzg_family` is used in the test? Would perhaps `kzg_proof_test` work instead?

tachyon/crypto/commitments/polynomial_openings_unittest.cc
chokobole: Actually #221 fixes the bug so missing changes looks more correct to me.

tachyon/crypto/commitments/kzg/BUILD.bazel
chokobole: I see, but what I meant was this are something like base test class for both `SHPlonkTest` and `GWCTest`

```c++
class SHPlonkTest : public KZGFamilyTest<PCS> {};
```

```c++
class GWCTest : public KZGFamilyTest<PCS> {};
```

chokobole: If it still makes you confused, I'd like to suggest `KZGTest` instead of `KZGProofTest` considering the name of the directory.

TomTaehoonKim (approved):
LGTM

Insun35 (approved):
LGTM

tachyon/crypto/commitments/kzg/BUILD.bazel
chokobole: I understand your point, but the impact of this naming isn't significant. So, why don't we rename it if we come up with a better idea later?

ashjeong: Ok sounds good~

ashjeong (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #333, opened 2024-02-27
perf: use `pdqsort` in production code for better performance

chokobole:
# Description

After reviewing the benchmarks for each circuit in the optimizations.md file, permutation was observed to be slower than Halo2. One of the main differences between Tachyon and Halo2 is the sorting algorithm used. Rust uses the stable sort "timsort", while C++ uses the unstable sort "quicksort". Timsort performs best when most of the input is already sorted.
Since we cannot make this assumption for the input expressions, we opt to use `pdqsort`, which provides better performance than quicksort.

TA raised this issue when implementing synthesizing circuits. `pdqsort` is now employed in every situation that requires a sort for a production purpose.


dongchangYoo (approved):
LGTM

ashjeong:
Please reflect my PR description changes in the commit message body.

ashjeong (approved):
LGTM

PR #341, opened 2024-03-07
fix(zk): fix number of affine points to be normalized

chokobole:
# Description

The number should have been `l_g1_query_jacobian.size()`, which represents the number of affine points to be normalized. However, occasionally it was set to 2, this PR fixes the bug.


dongchangYoo (approved):
LGTM

TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

PR #371, opened 2024-04-02
perf(crypto): avoid copying twice when returning a polynomial

chokobole:
# Description

The need for this optimization stems from the internal copying that occurs during multiplication and division between the polynomials. With `Poly& * operand`, copying occurs twice, so we use `Poly * operand` instead so that copying occurs just once.

```c++
Poly Func() {
  Poly& poly;
  Poly operand;
  return poly *= operand; // copying twice
}
```

```c++
Poly Func() {
  Poly& poly;
  Poly operand;
  return poly * operand; // copying once
}
```


dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

PR #372, opened 2024-04-02
perf: avoid in place operator where ineffective

chokobole:
# Description

This commit enhances performance by utilizing the `c = a op b` style instead of `c = a; c inplace-op b`. While the former style is nearly as performant as the latter when `a` is movable, it becomes inefficient when operating on prime fields or points on elliptic curves that are not able to be moved. This change ensures better performance and readability, reducing 2 lines to 1. Note that irrelevant changes may occur due to this adjustment.

### DIFF
--- a/tachyon/math/elliptic_curves/short_weierstrass/projective_point_impl.h
+++ b/tachyon/math/elliptic_curves/short_weierstrass/projective_point_impl.h
@@ -22,25 +22,20 @@ constexpr CLASS& CLASS::AddInPlace(const ProjectivePoint& other) {
 
   // http://www.hyperelliptic.org/EFD/g1p/auto-shortw-projective.html#addition-add-1998-cmo-2
   // Y1Z2 = Y1 * Z2
-  BaseField y1z2 = y_;
-  y1z2 *= other.z_;
+  BaseField y1z2 = y_ * other.z_;
 
   // X1Z2 = X1 * Z2
-  BaseField x1z2 = x_;
-  x1z2 *= other.z_;
+  BaseField x1z2 = x_ * other.z_;
 
   // Z1Z2 = Z1 * Z2
-  BaseField z1z2 = z_;
-  z1z2 *= other.z_;
+  BaseField z1z2 = z_ * other.z_;
 
   // u = Y2 * Z1 - Y1Z2
-  BaseField u = other.y_;
-  u *= z_;
+  BaseField u = other.y_ * z_;
   u -= y1z2;
 
   // v = X2 * Z1 - X1Z2
-  BaseField v = other.x_;
-  v *= z_;
+  BaseField v = other.x_ * z_;
   v -= x1z2;
 
   if (u.IsZero() && v.IsZero()) {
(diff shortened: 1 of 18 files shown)


dongchangYoo (approved):
LGTM

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

tachyon/math/elliptic_curves/short_weierstrass/projective_point_impl.h
```diff
@@ -155,60 +135,48 @@ constexpr CLASS& CLASS::DoubleInPlace() {
...
-  BaseField zz = z_;
-  zz.SquareInPlace();
+  BaseField zz = z_.Square();
 
   // w = a * ZZ + 3 * XX
-  BaseField w = xx;
-  w += w.Double();
+  BaseField w = xx + xx.Double();
   if constexpr (!Curve::Config::kAIsZero) {
     w += Curve::Config::MulByA(zz);
```
ashjeong, line 146:
```suggestion
  BaseField w = xx.Double();
  w += xx;
  if constexpr (!Curve::Config::kAIsZero) {
    w += Curve::Config::MulByA(std::move(zz));

File: tachyon/math/finite_fields/fp12.h [hunk 2]
```diff
@@ -282,8 +279,7 @@ class Fp12 final : public QuadraticExtensionField<Fp12<Config>> {
     this->c1_ -= b;
     // c0 = ((α₃β₃ + α₅β₄q) + (α₄β₃ + α₃β₄)x + (α₅β₃ + α₄β₄)x²)p +
     //      α₀β₀ + α₁β₀x + α₂β₀x², where p = y² and q = x³
-    this->c0_ = Config::MulByNonResidue(b);
-    this->c0_ += a;
+    this->c0_ = Config::MulByNonResidue(b) + a;
```
ashjeong, line 282: I don't think this should be changed:
its equal to the case 
```c++
z = b * c + a
```
which should be treated as such:
```c++
z = b * c;
z += a;
```

 REF[0] for future similar scenarios

tachyon/math/finite_fields/fp12.h
```diff
@@ -342,8 +335,7 @@ class Fp12 final : public QuadraticExtensionField<Fp12<Config>> {
     // c0 = (α₅β₄q + α₃β₄x + α₄β₄x²)p +
     //      (α₀β₀ + α₂β₁q) + (α₁β₀ + α₀β₁)x + (α₂β₀ + α₁β₁)x²,
     //      where p = y² and q = x³
-    this->c0_ = Config::MulByNonResidue(b);
-    this->c0_ += a;
+    this->c0_ = Config::MulByNonResidue(b) + a;
```
ashjeong, line 338: refer to REF[0]

ashjeong (changes requested):
I tried to only include small fixes/ changes related to the commit, yet there appears to be a lot of redundant or unneeded variables or calculations. Perhaps a separate pr may be needed to fix these.

tachyon/math/finite_fields/fp6.h
```diff
@@ -315,29 +315,15 @@ class Fp6<Config, std::enable_if_t<Config::kDegreeOverBaseField == 3>> final
...
-    this->c0_ = std::move(t0);
-    // c1 = α₀β₁
-    this->c1_ = std::move(t1);
     // c2 = α₁β₁
-    this->c2_ = std::move(t2);
+    this->c2_ = this->c1_ * beta1;
+    // c1 = α₀β₁
+    this->c1_ = this->c0_ * beta1;
+    // c0 = α₂β₁q
+    this->c0_ = Config::MulByNonResidue(t0);
```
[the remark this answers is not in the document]
chokobole: We can't do that, `c0` is used to compute `c1`, and `c1` is used to compute `c2` so we need to do in this order.

tachyon/math/elliptic_curves/short_weierstrass/projective_point_impl.h
chokobole: Actually `MulByA()` takes const reference as an argument!

tachyon/math/finite_fields/fp6.h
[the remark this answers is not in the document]
ashjeong: oops yea didn't catch that

PR #374, opened 2024-04-05
test(zk): enable `SelectorReplacerTest.Selector` test on optimization mode

chokobole:
# Description

`ReplaceSelectors()` only crashes only when compiled with debug mode, so this commit enables it using `EXPECT_DEBUG_DEATH`.

```c++
template <typename F>
class SelectorsReplacer : public Evaluator<F, std::unique_ptr<Expression<F>>> {
  std::unique_ptr<Expression<F>> Evaluate(const Expression<F>* input) override {

          if (selector.is_simple()) {
            LOG(DFATAL) << "Simple selector found in lookup argument";
          }
  }
```

### DIFF
--- a/tachyon/zk/expressions/evaluator/selector_replacer_unittest.cc
+++ b/tachyon/zk/expressions/evaluator/selector_replacer_unittest.cc
@@ -29,7 +29,7 @@ TEST_F(SelectorReplacerTest, Selector) {
   replacements.emplace_back(owned_replacements[0].get());
   replacements.emplace_back(owned_replacements[1].get());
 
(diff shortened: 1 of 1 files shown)


Insun35 (approved):
LGTM

ashjeong:
This PR fixes the issue in which running `bazel test --config linux //...` fails the test `//tachyon/zk/expressions/evaluator:expression_unittests`

ashjeong (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #376, opened 2024-04-05
fix: change compile option `maxvx512f` to `mavx2` for goldilocks

chokobole:
# Description

This is to resolve a runtime error of Goldilocks when building with g++-12 and optimization mode.

Plus, if we want to build with avx512, we need to add `defines = ["__AVX512__"]`, too.

See
https://github.com/0xPolygonHermez/zkevm-prover/blob/93c36f120c3201bfaf12b8f6f2796f53e72193ff/Makefile#L31-L37 for more details.


ashjeong (approved):
LGTM

Insun35 (approved):
LGTM

dongchangYoo (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #379, opened 2024-04-06
perf: avoid copying on semigroup operations

chokobole:
# Description

This PR avoids copying on semigroup operations, which are `Add()`, `Mul()`, `Double()` and `Square()`.

### DIFF
--- a/tachyon/math/base/semigroups.h
+++ b/tachyon/math/base/semigroups.h
@@ -67,7 +67,7 @@ namespace internal {
 SUPPORTS_BINARY_OPERATOR(Mul);
 SUPPORTS_UNARY_OPERATOR(DoSquare);
 SUPPORTS_BINARY_OPERATOR(Add);
-SUPPORTS_UNARY_IN_PLACE_OPERATOR(Double);
+SUPPORTS_UNARY_OPERATOR(DoDouble);
 
 template <typename T, typename = void>
 struct SupportsToBigInt : std::false_type {};
@@ -256,11 +256,23 @@ class AdditiveSemigroup {
 
   // a.Double(): 2a
   [[nodiscard]] constexpr auto Double() const {
-    if constexpr (internal::SupportsDoubleInPlace<G>::value) {
-      G g = *static_cast<const G*>(this);
-      return g.DoubleInPlace();
+    if constexpr (internal::SupportsDoDouble<G>::value) {
+      const G* g = static_cast<const G*>(this);
+      return g->DoDouble();
     } else {
       return operator+(static_cast<const G&>(*this));
+    }
+  }
+
+  // a.DoubleInPlace(): a = 2a
+  constexpr G& DoubleInPlace() {
+    if constexpr (internal::SupportsDoDoubleInPlace<G>::value) {
+      G* g = static_cast<G*>(this);
+      return g->DoDoubleInPlace();
+    } else if constexpr (internal::SupportsAddInPlace<G, G>::value) {
+      return operator+=(static_cast<const G&>(*this));
+    } else {
+      static_assert(base::AlwaysFalse<G>);
     }
   }
 
@@ -419,7 +431,8 @@ class AdditiveSemigroup {
     auto it = BitIteratorBE<BigInt<N>>::begin(&scalar, true);
     auto end = BitIteratorBE<BigInt<N>>::end(&scalar);
     while (it != end) {
-      if constexpr (internal::SupportsDoubleInPlace<G>::value) {
+      if constexpr (internal::SupportsDoDoubleInPlace<G>::value ||
+                    internal::SupportsAddInPlace<G, G>::value) {
         ret.DoubleInPlace();
       } else {
         ret = ret.Double();
--- a/tachyon/math/base/semigroups.h
+++ b/tachyon/math/base/semigroups.h
@@ -158,10 +158,8 @@ class MultiplicativeSemigroup {
   [[nodiscard]] constexpr auto Pow(const Scalar& scalar) const {
     if constexpr (std::is_constructible_v<BigInt<1>, Scalar>) {
       return DoPow(BigInt<1>(scalar));
-    } else if constexpr (internal::SupportsToBigInt<Scalar>::value) {
-      return DoPow(scalar.ToBigInt());
     } else {
-      static_assert(base::AlwaysFalse<G>);
+      return DoPow(scalar.ToBigInt());
     }
   }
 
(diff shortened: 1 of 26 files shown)


tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
```diff
@@ -522,11 +522,10 @@ class UnivariatePolynomialOp<UnivariateSparseCoefficients<F, MaxDegree>> {
   }
 
   static UnivariatePolynomial<D> ToDense(const UnivariatePolynomial<S>& self) {
-    std::vector<F> coefficients;
     size_t size = self.Degree() + 1;
-    coefficients.reserve(size);
-    for (size_t i = 0; i < size; ++i) {
-      coefficients.push_back(self[i]);
+    std::vector<F> coefficients(size);
```
dongchangYoo, line 584: We don't have to set every element of `coefficients` to default value.
``` diff
-     std::vector<F> coefficients(size);
+     std::vector<F> coefficients;
```

tachyon/math/polynomials/univariate/univariate_evaluations_ops.h
```diff
@@ -21,6 +21,24 @@ class UnivariateEvaluationsOp {
...
+    const std::vector<F>& r_evaluations = other.evaluations_;
+    if (l_evaluations.empty()) {
+      // 0 + g(x)
+      return other;
+    }
+    if (r_evaluations.empty()) {
+      // f(x) + 0
+      return self;
+    }
+    std::vector<F> o_evaluations(r_evaluations.size());
```
dongchangYoo, line 35:
```suggestion
    std::vector<F> o_evaluations;
    o_evaluations.reserve();
```

dongchangYoo (approved):
LGTM

ashjeong (commented):
change 4 commits:
Avoid copying twice **on** …
->
Avoid copying twice **in** …

Can the commit https://github.com/kroma-network/tachyon/pull/379/commits/4db1ac3d97d8543b03e798348b4e5f6daac1e3d9 be moved to before “avoid copying twice on addition” for continuity?

https://github.com/kroma-network/tachyon/pull/379/commits/a3b6bb398b57b97896fa42e12181aa43694b0ec9
refac(math): refactor fast **retrun** case of polynomial operations
->
refac(math): refactor fast **return** case of polynomial operations

Rather than a refactor(changing structure of code, but no logic changes) this seems more like chore, since logic is changed.

tachyon/math/base/semigroups.h
ashjeong, line 1: lines 95-106
could we change to 
```c++
template <
      typename G2,
      std::enable_if_t<internal::SupportsMul<G, G2>::value>* = nullptr>
  constexpr auto operator*(const G2& other) const {
      const G* g = static_cast<const G*>(this);
      return g->Mul(other);
  }
  ```
& same for add?

File: tachyon/math/base/semigroups.h [hunk 3]
```diff
@@ -268,14 +268,26 @@ class AdditiveSemigroup {
...
+      const G* g = static_cast<const G*>(this);
+      return g->DoDouble();
     } else {
       return operator+(static_cast<const G&>(*this));
     }
   }
 
+  // a.DoubleInPlace(): a = 2a
+  constexpr G& DoubleInPlace() {
+    if constexpr (internal::SupportsDoDoubleInPlace<G>::value) {
```
ashjeong, line 281: Won't double always be supported?

tachyon/math/finite_fields/cubic_extension_field.h
```diff
@@ -187,47 +187,26 @@ class CubicExtensionField : public CyclotomicMultiplicativeSubgroup<Derived> {
...
-    c1_ = x + Config::MulByNonResidue(v2);
-    // c2 = c0 * other.c2 + c2 * other.c0 - c1 * other.c1
-    c2_ = y + v1;
+  constexpr Derived& MulInPlace(const Derived& other) {
+    DoMul(*static_cast<const Derived*>(this), other,
+          *static_cast<Derived*>(this));
     return *static_cast<Derived*>(this);
   }
 
+  constexpr Derived Mul(const BaseField& element) const {
```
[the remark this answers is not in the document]
chokobole: Then this would be better to be done by the different PR, since there will be a lot of cases.

tachyon/math/base/semigroups.h
chokobole: Oh you're right! you mean we have to remove `internal::SupportsMulInPlace<G, G2>::value>` part right?

tachyon/math/base/semigroups.h
chokobole: Yes, `Double()` always be supported!

tachyon/math/base/semigroups.h
ashjeong: yep!

tachyon/math/base/semigroups.h
```diff
@@ -92,17 +108,14 @@ class MultiplicativeSemigroup {
...
+            std::enable_if_t<internal::SupportsMul<G, G2>::value>* = nullptr>
   constexpr auto operator*(const G2& other) const {
     if constexpr (internal::SupportsMul<G, G2>::value) {
       const G* g = static_cast<const G*>(this);
       return g->Mul(other);
     } else {
-      G g = *static_cast<const G*>(this);
-      return g.MulInPlace(other);
+      static_assert(base::AlwaysFalse<G>);
     }
```
ashjeong, line 119: I think once `SupportsMulInPlace` is removed as an enable if condition, there's no reason to have this if/else statement (+ same for `Add`) 
```suggestion
    const G* g = static_cast<const G*>(this);
    return g->Mul(other);
```

ashjeong (approved):
LGTM

Insun35 (approved):
LGTM

tachyon/math/polynomials/multivariate/multilinear_extension_ops.h
```diff
@@ -16,6 +16,19 @@ class MultilinearExtensionOp<MultilinearDenseEvaluations<F, MaxDegree>> {
  public:
   using D = MultilinearDenseEvaluations<F, MaxDegree>;
 
+  static MultilinearExtension<D> Add(const MultilinearExtension<D>& self,
```
TomTaehoonKim, line 19: How about adding an early return here as well when self or other is zero?

TomTaehoonKim (changes requested):

tachyon/math/polynomials/univariate/univariate_evaluations_ops.h
```diff
@@ -68,6 +68,24 @@ class UnivariateEvaluationsOp {
...
+    const std::vector<F>& l_evaluations = self.evaluations_;
+    const std::vector<F>& r_evaluations = other.evaluations_;
+    if (l_evaluations.empty()) {
+      // 0 * g(x)
+      return Poly::Zero();
+    }
+    if (r_evaluations.empty()) {
+      // f(x) * 0
+      return Poly::Zero();
+    }
```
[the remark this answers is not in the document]
chokobole: Checking `IsOne()` of `UnivariateEvaluations()` is very expensive, that's why this only checks `xxx.empty()`!

tachyon/math/polynomials/multivariate/multilinear_extension_ops.h
chokobole: `IsZero()` checking possible is expensive, but I think `CHECK_EQ(l_evaluations.size(), r_evaluations.size());` seems a bug! I'll fix it!

TomTaehoonKim (approved):
LGTM

PR #385, opened 2024-04-07
build(math): generate m31 field as non fft friendly field

chokobole:
# Description

Since mersenne 31 has only 1 2-adicity, it should be treated as non-fft friendly.

```
M = 2³¹ - 1
M - 1 = 2³¹ - 2 = 2(2³⁰ - 1)
```

### DIFF
--- a/tachyon/math/finite_fields/generator/prime_field_generator/config.h.tpl
+++ b/tachyon/math/finite_fields/generator/prime_field_generator/config.h.tpl
@@ -51,15 +51,14 @@ class TACHYON_EXPORT %{class}Config {
     %{one_mont_form}
   });
 
-  constexpr static BigInt<%{n}> kSubgroupGenerator = BigInt<%{n}>({
-    %{subgroup_generator}
-  });
-
   constexpr static bool kHasTwoAdicRootOfUnity = %{has_two_adic_root_of_unity};
 
   constexpr static bool kHasLargeSubgroupRootOfUnity = %{has_large_subgroup_root_of_unity};
 
 %{if kHasTwoAdicRootOfUnity}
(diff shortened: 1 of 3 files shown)


tachyon/math/finite_fields/generator/prime_field_generator/config.h.tpl
```diff
@@ -51,15 +51,14 @@ class TACHYON_EXPORT %{class}Config {
...
-  });
-
   constexpr static bool kHasTwoAdicRootOfUnity = %{has_two_adic_root_of_unity};
 
   constexpr static bool kHasLargeSubgroupRootOfUnity = %{has_large_subgroup_root_of_unity};
 
 %{if kHasTwoAdicRootOfUnity}
+  constexpr static BigInt<%{n}> kSubgroupGenerator = BigInt<%{n}>({
+    %{subgroup_generator}
+  });
```
dongchangYoo, line 61: All prime fields have `kSubgroupGenerator` which generates every element of the field.
As i know, a generator which generates every element of 2-adic-group is called `kTwoAdicRootOfUnity` in our prime field config.

dongchangYoo: Therefore this commit should be reverted.

chokobole: I don't think so. I believe this value is only needed by fft friendly field

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #392, opened 2024-04-16
fix(math): add missing changes from #388

chokobole:
# Description

This PR addresses the compilation error that occurred in #388.

### DIFF
--- a/tachyon/math/base/big_int_unittest.cc
+++ b/tachyon/math/base/big_int_unittest.cc
@@ -31,6 +31,8 @@ TEST(BigIntTest, DecString) {
   // 1 << 65
   BigInt<2> big_int = *BigInt<2>::FromDecString("36893488147419103232");
   EXPECT_EQ(big_int.ToString(), "36893488147419103232");
+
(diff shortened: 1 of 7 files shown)


ashjeong (commented):
Should the same functions in `prime_field_gpu_debug.h` also be changed?

Insun35 (approved):
LGTM

chokobole:
> Should the same functions in `prime_field_gpu_debug.h` also be changed?

Yes, since this is not used anywhere, so I didn't notice that! Thanks :)

TomTaehoonKim (approved):
LGTM

TomTaehoonKim:
I just thought of it, maybe we should've added test cases for #388 where `std::nullopt` is returned when a malformed input is given. Perhaps this is too trivial?

chokobole:
> I just thought of it, maybe we should've added test cases for #388 where `std::nullopt` is returned when a malformed input is given. Perhaps this is too trivial?

I added it!

Insun35 (approved):
LGTM

chokobole:
> Should the same functions in `prime_field_gpu_debug.h` also be changed?

I changed it!

PR #394, opened 2024-04-19
refac(halo2): change concrete struct type to trait type for rng

k-gunjan:
# Description
This PR makes rng parameter of create_proof function generic over RngCore and RngState.
Defined trait RngState and implemented for XORShiftRng

### DIFF
--- a/vendors/halo2/src/xor_shift_rng.rs
+++ b/vendors/halo2/src/xor_shift_rng.rs
@@ -12,12 +12,16 @@ pub mod ffi {
     }
 }
 
+pub trait RngState {
+    fn state(&self) -> Vec<u8>;
+}
+
 pub struct XORShiftRng {
     inner: cxx::UniquePtr<ffi::XORShiftRng>,
 }
 
-impl XORShiftRng {
-    pub fn state(&self) -> Vec<u8> {
+impl RngState for XORShiftRng {
+    fn state(&self) -> Vec<u8> {
         self.inner.state()
     }
 }
@@ -63,9 +67,9 @@ impl rand_core::RngCore for XORShiftRng {
 
 #[cfg(test)]
 mod test {
-    use rand_core::{RngCore, SeedableRng};
-
+    use super::RngState;
     use crate::consts::SEED;
+    use rand_core::{RngCore, SeedableRng};
 
     #[test]
     fn test_rng() {
--- a/vendors/halo2/src/circuits/shuffle_circuit.rs
+++ b/vendors/halo2/src/circuits/shuffle_circuit.rs
@@ -315,7 +315,7 @@ mod test {
             };
             let mut transcript = TachyonBlake2bWrite::init(vec![]);
 
-            tachyon_create_proof::<_, _, _, _, _>(
+            tachyon_create_proof::<_, _, _, _, _, _>(
                 &mut prover,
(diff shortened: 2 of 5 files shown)


chokobole:
Thanks for your contribution! 🎉

Firstly, could you please follow the commit convention outlined [here](https://github.com/kroma-network/tachyon/blob/main/docs/how_to_contribute/conventions.md#commits)?

Additionally, it's essential that every commit is buildable and testable, but unfortunately, your PR doesn't meet these criteria. To address this, I suggest squashing all the commits into a single one to ensure they're buildable and testable.

Let me know if you have any questions or need further clarification. Thanks again for your effort! 👍

vendors/halo2/src/traits.rs
```diff
@@ -0,0 +1,3 @@
+pub trait RngState {
+    fn state(&self) -> Vec<u8>;
+}
```
chokobole, line 3: What about moving this trait to `xor_shift_rng.rs`?

k-gunjan: it can be shifted there. 
However i want to draw your attenton to a fact that other types except `XORShiftRng` in `xor_shift_rng.rs` might need to implement it, in that case keeping its definition separate would allow implementation independent of  `xor_shift_rng.rs` module.

let me know your view and confirm the changes.

chokobole: Oh I see. But in the case of Halo2, `XorShiftRng` is only used. Could we consider this approach when there should be more implementations to come? Since it looks not natural that `traits.rs` has only this trait even though there are other traits.

k-gunjan: Sounds cool. l'll move the trait definition to `xor_shift_rng.rs` itself .

vendors/halo2/src/xor_shift_rng.rs
```diff
@@ -63,9 +67,9 @@ impl rand_core::RngCore for XORShiftRng {
 
 #[cfg(test)]
 mod test {
-    use rand_core::{RngCore, SeedableRng};
-
+    use super::RngState;
     use crate::consts::SEED;
+    use rand_core::{RngCore, SeedableRng};
```
chokobole, line 72: Did you intend to move this statement here?

k-gunjan: Oh no, this moved by `cargo fmt` automatically while saving. Is it okay to leave it like this ?

chokobole: Yup normally we separate formatting as a style commit, but it's minor, so just leave it!

chokobole (approved):
LGTM

vendors/halo2/src/xor_shift_rng.rs
k-gunjan: got it.

ashjeong (commented):
Could you change your commit's scope to match the PR's scope (Halo2)? Thanks again for the contribution~ 🥰

Insun35 (approved):
LGTM

TomTaehoonKim (approved):
LGTM

chokobole:
Close this issue since we need this feature immediately to support different `ChaCha20` for https://github.com/scroll-tech/snark-verifier!

PR #418, opened 2024-05-13
build: use latest docker image

wo-o:
Use latest docker image


Insun35 (commented):
Please change your commit message to:
build: use latest docker image

PR #430, opened 2024-06-07
feat(math): impl remove zeros for sparse univariate poly

batzor:
# Description

- Change `RemoveHighDegreeZeros()` to `RemoveZeros()` in univariate spare polynomial
- Enforce clean construction of polynomials. After this change, all poly operations are expected to operate on _clean_ polynomials and output _clean_ polynomials.

### DIFF
--- a/tachyon/math/polynomials/univariate/lagrange_interpolation.h
+++ b/tachyon/math/polynomials/univariate/lagrange_interpolation.h
@@ -31,7 +31,7 @@ bool LagrangeInterpolate(
   }
 
   if (points.size() == 1) {
-    *ret = Poly(Coeffs({evals[0]}));
+    *ret = Poly(Coeffs({evals[0]}, true));
     return true;
   }
 
@@ -101,7 +101,7 @@ bool LagrangeInterpolate(
       coeffs_sums[0][j] += coeffs_sums[i][j];
     }
   }
-  *ret = Poly(Coeffs(std::move(coeffs_sums[0])));
+  *ret = Poly(Coeffs(std::move(coeffs_sums[0]), true));
   return true;
 }
 
--- a/tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
+++ b/tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
@@ -71,8 +71,12 @@ class UnivariateDenseCoefficients {
   }
 
   constexpr static UnivariateDenseCoefficients Random(size_t degree) {
-    return UnivariateDenseCoefficients(
-        base::CreateVector(degree + 1, []() { return F::Random(); }));
+    std::vector v =
+        base::CreateVector(degree + 1, []() { return F::Random(); });
+    if (v[degree].IsZero()) {
+      v[degree] = F::One();
+    }
+    return UnivariateDenseCoefficients(std::move(v));
   }
 
   // Return dense coefficients according to the given |roots|.
--- a/tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
+++ b/tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
@@ -49,15 +49,17 @@ class UnivariateDenseCoefficients {
 
   constexpr UnivariateDenseCoefficients() = default;
   constexpr explicit UnivariateDenseCoefficients(
-      const std::vector<F>& coefficients)
+      const std::vector<F>& coefficients, bool cleanup = false)
       : coefficients_(coefficients) {
+    if (cleanup) RemoveHighDegreeZeros();
     CHECK_LE(Degree(), kMaxDegree);
-    RemoveHighDegreeZeros();
   }
-  constexpr explicit UnivariateDenseCoefficients(std::vector<F>&& coefficients)
+
+  constexpr explicit UnivariateDenseCoefficients(std::vector<F>&& coefficients,
+                                                 bool cleanup = false)
       : coefficients_(std::move(coefficients)) {
+    if (cleanup) RemoveHighDegreeZeros();
     CHECK_LE(Degree(), kMaxDegree);
-    RemoveHighDegreeZeros();
   }
 
   constexpr static UnivariateDenseCoefficients Zero() {
@@ -189,7 +191,7 @@ class UnivariateDenseCoefficients {
         coefficients[size >> 1] *= r;
       }
     }
-    return UnivariateDenseCoefficients(std::move(coefficients));
+    return UnivariateDenseCoefficients(std::move(coefficients), true);
   }
 
   std::string ToString() const {
@@ -213,6 +215,20 @@ class UnivariateDenseCoefficients {
       }
     }
     return ss.str();
+  }
+
+  void RemoveHighDegreeZeros() {
+    while (!IsZero()) {
+      if (coefficients_.back().IsZero()) {
+        coefficients_.pop_back();
+      } else {
+        break;
+      }
+    }
+  }
+
+  constexpr bool IsClean() const {
+    return coefficients_.size() == 0 || !coefficients_.back().IsZero();
   }
 
  private:
@@ -261,16 +277,6 @@ class UnivariateDenseCoefficients {
                              result *= point;
                              return result += coeff;
                            });
-  }
-
-  void RemoveHighDegreeZeros() {
-    while (!IsZero()) {
-      if (coefficients_.back().IsZero()) {
-        coefficients_.pop_back();
-      } else {
-        break;
-      }
-    }
   }
 
   std::vector<F> coefficients_;
--- a/tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
+++ b/tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
@@ -766,6 +766,9 @@ class UnivariatePolynomialOp<UnivariateSparseCoefficients<F, MaxDegree>> {
   }
 
   static UnivariatePolynomial<D> ToDense(const UnivariatePolynomial<S>& self) {
+    if (self.IsZero()) {
+      return UnivariatePolynomial<D>::Zero();
+    }
     size_t size = self.Degree() + 1;
     std::vector<F> coefficients(size);
     OPENMP_PARALLEL_FOR(size_t i = 0; i < size; ++i) {
--- a/tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
+++ b/tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
@@ -835,7 +835,6 @@ class UnivariatePolynomialOp<UnivariateSparseCoefficients<F, MaxDegree>> {
     }
 
     c.coefficients_ = S(std::move(c_terms));
-    c.coefficients_.RemoveHighDegreeZeros();
   }
 
   static void DoMul(const UnivariatePolynomial<S>& a,
@@ -864,7 +863,6 @@ class UnivariatePolynomialOp<UnivariateSparseCoefficients<F, MaxDegree>> {
     }
     pdqsort(c_terms.begin(), c_terms.end());
     c.coefficients_ = S(std::move(c_terms));
-    c.coefficients_.RemoveHighDegreeZeros();
   }
 };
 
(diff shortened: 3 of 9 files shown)


tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
```diff
@@ -104,13 +104,13 @@ class UnivariateSparseCoefficients {
   constexpr explicit UnivariateSparseCoefficients(
       const std::vector<Term>& terms)
       : terms_(terms) {
+    RemoveZeros();
     CHECK_LE(Degree(), kMaxDegree);
-    RemoveHighDegreeZeros();
```
chokobole, line 108: Why did you switch the order?

batzor: It will handle cases when the highest degree terms are 0.

batzor: To rephrase, Degree() can become less that kMaxDegree after RemoveZeros()

tachyon/math/polynomials/univariate/univariate_sparse_polynomial_unittest.cc
```diff
@@ -324,6 +323,7 @@ TEST_F(UnivariateSparsePolynomialTest, MulScalar) {
   }
 
   Poly actual = poly * scalar;
+  actual.coefficients().RemoveZeros();
```
TomTaehoonKim, line 326: Isn't `RemoveZeros()` called when multiplication is done? Do we need this code?

TomTaehoonKim (changes requested):

dongchangYoo (approved):
LGTM

tachyon/math/polynomials/univariate/univariate_sparse_polynomial_unittest.cc
batzor: This test was failing in some random cases and I found that it fails when scalar is 0 the expected is constructed with ill-formed terms and cause the test to fail. I changed to following:

`expected.coefficients().RemoveZeros();`

Insun35 (approved):
LGTM

tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
```diff
@@ -69,8 +71,11 @@ class UnivariateDenseCoefficients {
   }
 
   constexpr static UnivariateDenseCoefficients Random(size_t degree) {
-    return UnivariateDenseCoefficients(
-        base::CreateVector(degree + 1, []() { return F::Random(); }));
+    auto v = base::CreateVector(degree + 1, []() { return F::Random(); });
```
ashjeong, line 74: Could you explicitly name the type as `std::vector<F>`?
```suggestion
    std::vector<F> v = base::CreateVector(degree + 1, []() { return F::Random(); });

**Review by ashjeong — CHANGES_REQUESTED** [review]

**Review by ashjeong — APPROVED** [review]
LGTM

File: tachyon/math/polynomials/univariate/univariate_dense_coefficients.h [hunk 4]
```diff
@@ -69,8 +71,12 @@ class UnivariateDenseCoefficients {
...
 
   constexpr static UnivariateDenseCoefficients Random(size_t degree) {
-    return UnivariateDenseCoefficients(
-        base::CreateVector(degree + 1, []() { return F::Random(); }));
+    std::vector v =
+        base::CreateVector(degree + 1, []() { return F::Random(); });
+    if (v[degree].IsZero()) {
+      v[degree] = F::One();
+    }
+    return UnivariateDenseCoefficients(std::move(v));
```
TomTaehoonKim, line 79: Please make this change as a separate commit.

TomTaehoonKim (changes requested):

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
```diff
@@ -102,15 +102,16 @@ class UnivariateSparseCoefficients {
 
   constexpr UnivariateSparseCoefficients() = default;
   constexpr explicit UnivariateSparseCoefficients(
-      const std::vector<Term>& terms)
+      const std::vector<Term>& terms, bool cleanup = false)
       : terms_(terms) {
+    if (cleanup) RemoveZeros();
     CHECK_LE(Degree(), kMaxDegree);
-    RemoveHighDegreeZeros();
```
TomTaehoonKim, line 108: This should've been removed in the previous commit.

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
```diff
@@ -102,15 +102,16 @@ class UnivariateSparseCoefficients {
...
   }
-  constexpr explicit UnivariateSparseCoefficients(std::vector<Term>&& terms)
+  constexpr explicit UnivariateSparseCoefficients(std::vector<Term>&& terms,
+                                                  bool cleanup = false)
       : terms_(std::move(terms)) {
+    if (cleanup) RemoveZeros();
     CHECK_LE(Degree(), kMaxDegree);
-    RemoveHighDegreeZeros();
```
TomTaehoonKim, line 113: This should've been removed in the previous commit.

TomTaehoonKim:
@batzor For the second commit, I can agree for the sparse coefficients but not with the dense coefficients. What's the motivation behind reducing the usage of remove high degree zeros for dense coefficients?

tachyon/math/polynomials/univariate/univariate_sparse_polynomial_unittest.cc
```diff
@@ -40,11 +40,7 @@ TEST_F(UnivariateSparsePolynomialTest, IsZero) {
   EXPECT_TRUE(Poly().IsZero());
   EXPECT_TRUE(Poly::Zero().IsZero());
 
-  Poly p = Poly(Coeffs({{0, GF7(0)}, {1, GF7(0)}}));
-  p.coefficients().RemoveZeros();
-  EXPECT_TRUE(p.IsZero());
```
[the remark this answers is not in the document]
batzor: With this PR, we are enforcing polynomial construction with only _cleaned up_ coefficients. So this test fails in line 43 because the constructor checks if the coefficients are clean.

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
batzor: This is the first commit :)

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
batzor: This is in the first commit as well

tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
batzor: Umm may I ask why? I think it is part of enforcing proper polynomial construction. It wouldn't be a bug without this refactoring.

batzor:
@TomTaehoonKim Indeed, the `RemoveHighDegreeZeros()` is low cost operation so it's not that necessary but I just did the change for the sake of consistency and clarity.

tachyon/math/polynomials/univariate/univariate_sparse_polynomial_unittest.cc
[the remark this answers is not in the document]
batzor: I reverted the removal and fixed the test.

tachyon/math/polynomials/univariate/univariate_polynomial_ops.h
```diff
@@ -766,6 +766,9 @@ class UnivariatePolynomialOp<UnivariateSparseCoefficients<F, MaxDegree>> {
   }
 
   static UnivariatePolynomial<D> ToDense(const UnivariatePolynomial<S>& self) {
+    if (self.IsZero()) {
+      return UnivariatePolynomial<D>::Zero();
+    }
```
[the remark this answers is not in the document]
batzor: Done :)

tachyon/math/polynomials/univariate/univariate_dense_coefficients.h
batzor: Done :)

tachyon/math/polynomials/univariate/lagrange_interpolation.h
```diff
@@ -101,7 +101,7 @@ bool LagrangeInterpolate(
       coeffs_sums[0][j] += coeffs_sums[i][j];
     }
   }
-  *ret = Poly(Coeffs(std::move(coeffs_sums[0])));
+  *ret = Poly(Coeffs(std::move(coeffs_sums[0]), true));
```
TomTaehoonKim, line 104: What about cases like
1. https://github.com/kroma-network/tachyon/blob/e182278d97c58c934c3d81f00aa710df9c1c5115/tachyon/math/polynomials/univariate/lagrange_interpolation.h#L34
and
2. https://github.com/kroma-network/tachyon/blob/e182278d97c58c934c3d81f00aa710df9c1c5115/tachyon/zk/plonk/vanishing/vanishing_prover_impl.h#L178
?
I'm still not really convinced with "enforcing clean construction of polynomials.

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
```diff
@@ -289,16 +294,6 @@ class UnivariateSparseCoefficients {
     return &it->coefficient;
   }
 
-  void RemoveHighDegreeZeros() {  // Fix to RemoveZeros
```
TomTaehoonKim, line 292: Have you tried build and tests? I'm pretty sure by removing this, it will fail because we call this when doing polynomial operations. https://github.com/kroma-network/tachyon/blob/e182278d97c58c934c3d81f00aa710df9c1c5115/tachyon/math/polynomials/univariate/univariate_polynomial_ops.h#L835

batzor: Yes I did build, test and lint.

Umm that seems to be from old commit? That line is removed in this commit if you look above.

tachyon/math/polynomials/univariate/lagrange_interpolation.h
batzor: As I understood, some operations rely on input polynomial to be _clean_. 
For example, https://github.com/kroma-network/tachyon/blob/e182278d97c58c934c3d81f00aa710df9c1c5115/tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h#L227-L234

I think enforcing it will be more bug-free in the future. As for the cases you mentioned, I could recheck everything and pass `true` in the needed parts. I thought of making the dense polynomial to cleanup by default but still in your second case, we will have to pass `true` since the polynomial can be sparse.

batzor:
@chokobole I reordered the commits as you requested. Hope its more clear now;;

tachyon/math/polynomials/univariate/lagrange_interpolation.h
batzor: For now I added `true` to the cases you mentioned. I couldn't find any other instance.

tachyon/math/polynomials/univariate/univariate_sparse_coefficients.h
```diff
@@ -283,6 +283,13 @@ class UnivariateSparseCoefficients {
     return ss.str();
   }
 
+  void RemoveZeros() {
+    terms_.erase(
+        std::remove_if(terms_.begin(), terms_.end(),
+                       [](const Term& x) { return x.coefficient.IsZero(); }),
+        terms_.end());
```
chokobole, line 290:
```suggestion
    base::EraseIf(terms_, [](const Term& x) { return x.coefficient.IsZero(); }));
```

batzor: done!

chokobole (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #453, opened 2024-06-27
feat(zk): implement air constraint system

batzor:
# Description

This PR implements the Plonky3 AIR constraint system.

### DIFF
--- a/tachyon/zk/air/constraint_system/BUILD.bazel
+++ b/tachyon/zk/air/constraint_system/BUILD.bazel
@@ -1,6 +1,16 @@
-load("//bazel:tachyon_cc.bzl", "tachyon_cc_library")
+load("//bazel:tachyon_cc.bzl", "tachyon_cc_library", "tachyon_cc_unittest")
 
 package(default_visibility = ["//visibility:public"])
+
+tachyon_cc_library(
+    name = "constraint_system",
+    hdrs = ["constraint_system.h"],
+    deps = [
+        ":variable",
+        "//tachyon/math/matrix:matrix_types",
+        "//tachyon/zk/air/expressions:air_evaluator",
+    ],
+)
 
 tachyon_cc_library(
     name = "variable",
@@ -9,5 +19,16 @@ tachyon_cc_library(
         "//tachyon:export",
         "//tachyon/base:logging",
         "@com_google_absl//absl/strings",
+    ],
+)
+
+tachyon_cc_unittest(
+    name = "air_unittests",
+    srcs = ["constraint_system_unittest.cc"],
+    deps = [
+        ":constraint_system",
+        "//tachyon/math/finite_fields/test:finite_field_test",
+        "//tachyon/math/finite_fields/test:gf7",
+        "//tachyon/zk/air/expressions:expression_factory",
     ],
 )
--- a/tachyon/zk/air/constraint_system/BUILD.bazel
+++ b/tachyon/zk/air/constraint_system/BUILD.bazel
@@ -1,0 +1,13 @@
+load("//bazel:tachyon_cc.bzl", "tachyon_cc_library")
+
+package(default_visibility = ["//visibility:public"])
+
+tachyon_cc_library(
+    name = "variable",
+    hdrs = ["variable.h"],
+    deps = [
+        "//tachyon:export",
+        "//tachyon/base:logging",
+        "@com_google_absl//absl/strings",
+    ],
+)
--- a/tachyon/zk/air/constraint_system/constraint_system.h
+++ b/tachyon/zk/air/constraint_system/constraint_system.h
@@ -1,0 +1,95 @@
+#ifndef TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
+#define TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
+
+#include <stddef.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "tachyon/math/matrix/matrix_types.h"
+#include "tachyon/zk/air/constraint_system/variable.h"
+#include "tachyon/zk/air/expressions/air_evaluator.h"
+
+namespace tachyon::zk::air {
+
+template <typename F>
+class ConstraintSystem {
+ public:
+  using RowMajorMatrix = math::RowMajorMatrix<F>;
+  using ConstMatrixBlock = Eigen::Block<const RowMajorMatrix>;
+
+  ConstraintSystem() = default;
+  ConstraintSystem(size_t num_public_values, size_t main_width,
+                   size_t preprocessed_width)
+      : num_public_values_(num_public_values),
+        main_width_(main_width),
+        preprocessed_width_(preprocessed_width),
+        preprocessed_buf_(std::make_unique<char[]>(sizeof(ConstMatrixBlock))) {}
+
+  void EnforceFirstRowConstraint(std::unique_ptr<Expression<F>> constraint) {
+    EnforceConstraint(ExpressionFactory<F>::FirstRow(std::move(constraint)));
+  }
+
+  void EnforceLastRowConstraint(std::unique_ptr<Expression<F>> constraint) {
+    EnforceConstraint(ExpressionFactory<F>::LastRow(std::move(constraint)));
+  }
+
+  void EnforceTransitionConstraint(std::unique_ptr<Expression<F>> constraint) {
+    EnforceConstraint(ExpressionFactory<F>::Transition(std::move(constraint)));
+  }
+
+  void EnforceConstraint(std::unique_ptr<Expression<F>> constraint) {
+    constraints_.push_back(std::move(constraint));
+  }
+
+  bool IsSatisfied(AirEvaluator<F>& evaluator,
+                   const std::vector<F>& public_values,
+                   const RowMajorMatrix& main,
+                   const RowMajorMatrix* preprocessed = nullptr) const {
+    CHECK_EQ(public_values.size(), num_public_values_);
+    CHECK_EQ(main.cols(), main_width_);
+    CHECK_GE(main.rows(), 1);
+    if (preprocessed) {
+      CHECK_EQ(preprocessed->cols(), preprocessed_width_);
+      CHECK_EQ(preprocessed->rows(), main.rows());
+    } else {
+      CHECK_EQ(preprocessed_width_, 0);
+    }
+
+    evaluator.set_public_values(public_values);
+    evaluator.set_num_rows(main.rows());
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
+      const ConstMatrixBlock main_window(main, i, 0, 2, main_width_);
+
+      if (preprocessed) {
+        new (preprocessed_buf_.get())
+            ConstMatrixBlock(*preprocessed, i, 0, 2, preprocessed_width_);
+        evaluator.SetCurrentWindowData(
+            i, &main_window,
+            reinterpret_cast<const ConstMatrixBlock*>(preprocessed_buf_.get()));
+      } else {
+        evaluator.SetCurrentWindowData(i, &main_window, nullptr);
+      }
+
+      for (const std::unique_ptr<Expression<F>>& constraint : constraints_) {
+        if (!evaluator.Evaluate(constraint.get()).IsZero()) {
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+ private:
+  const size_t num_public_values_ = 0;
+  const Eigen::Index main_width_ = 0;
+  const Eigen::Index preprocessed_width_ = 0;
+  std::vector<std::unique_ptr<Expression<F>>> constraints_;
+  std::unique_ptr<char[]> preprocessed_buf_;
+};
+
+}  // namespace tachyon::zk::air
+
+#endif  // TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
--- a/tachyon/zk/air/constraint_system/constraint_system_unittest.cc
+++ b/tachyon/zk/air/constraint_system/constraint_system_unittest.cc
@@ -1,0 +1,108 @@
+#include "tachyon/zk/air/constraint_system/constraint_system.h"
+
+#include "gtest/gtest.h"
+
+#include "tachyon/math/finite_fields/test/finite_field_test.h"
+#include "tachyon/math/finite_fields/test/gf7.h"
+#include "tachyon/math/matrix/matrix_types.h"
+#include "tachyon/zk/air/expressions/air_evaluator.h"
+#include "tachyon/zk/air/expressions/expression_factory.h"
+
+namespace tachyon::zk::air {
+
+namespace {
+
+using F = math::GF7;
+
+class ConstraintSystemTest : public math::FiniteFieldTest<F> {};
+
+}  // namespace
+
+TEST_F(ConstraintSystemTest, FibonacciAirTest) {
+  ConstraintSystem<F> constraint_system(3, 2, 0);
+  // Check initial inputs in the first row: public₀ = main₀,₀
+  std::unique_ptr<Expression<F>> expr1 =
+      ExpressionFactory<F>::Variable(Variable::Public(0)) -
+      ExpressionFactory<F>::Variable(Variable::Main(0, 0));
+  constraint_system.EnforceFirstRowConstraint(std::move(expr1));
+
+  // Check initial inputs in the first row: public₁ = main₀,₁
+  std::unique_ptr<Expression<F>> expr2 =
+      ExpressionFactory<F>::Variable(Variable::Public(1)) -
+      ExpressionFactory<F>::Variable(Variable::Main(0, 1));
+  constraint_system.EnforceFirstRowConstraint(std::move(expr2));
+
+  // Check transition step: main₁,₀  = main₀,₁
+  std::unique_ptr<Expression<F>> expr3 =
+      ExpressionFactory<F>::Variable(Variable::Main(1, 0)) -
+      ExpressionFactory<F>::Variable(Variable::Main(0, 1));
+  constraint_system.EnforceTransitionConstraint(std::move(expr3));
+
+  // Check transition step: main₁,₁ = main₀,₀ + main₀,₁
+  std::unique_ptr<Expression<F>> expr4 =
+      ExpressionFactory<F>::Variable(Variable::Main(1, 1)) -
+      ExpressionFactory<F>::Variable(Variable::Main(0, 0)) -
+      ExpressionFactory<F>::Variable(Variable::Main(0, 1));
+  constraint_system.EnforceTransitionConstraint(std::move(expr4));
+
+  // In the last row, check if output is same: main₁,₁ = public₂
+  std::unique_ptr<Expression<F>> expr5 =
+      ExpressionFactory<F>::Variable(Variable::Main(1, 1)) -
+      ExpressionFactory<F>::Variable(Variable::Public(2));
+  constraint_system.EnforceLastRowConstraint(std::move(expr5));
+
+  std::vector<F> public_trace = {F(1), F(1), F(1)};
+  math::RowMajorMatrix<F> main_trace(5, 2);
+
+  // clang-format off
+  main_trace << F(1), F(1),
+                F(1), F(2),
+                F(2), F(3),
+                F(3), F(5),
+                F(5), F(1);
+  // clang-format on
+
+  AirEvaluator<F> evaluator;
+  bool is_satisfied = constraint_system.IsSatisfied(evaluator, public_trace,
+                                                    main_trace, nullptr);
+  EXPECT_TRUE(is_satisfied);
+}
+
+TEST_F(ConstraintSystemTest, PreprocessedAirTest) {
+  ConstraintSystem<F> constraint_system(0, 2, 2);
+
+  std::unique_ptr<Expression<F>> expr1 =
+      ExpressionFactory<F>::Variable(Variable::Preprocessed(0, 0)) *
+      (ExpressionFactory<F>::Variable(Variable::Main(0, 0)) -
+       ExpressionFactory<F>::Variable(Variable::Main(1, 0)));
+  constraint_system.EnforceTransitionConstraint(std::move(expr1));
+
+  std::unique_ptr<Expression<F>> expr2 =
+      ExpressionFactory<F>::Variable(Variable::Preprocessed(0, 1)) *
+      (ExpressionFactory<F>::Variable(Variable::Main(0, 1)) -
+       ExpressionFactory<F>::Variable(Variable::Main(1, 1)));
+  constraint_system.EnforceTransitionConstraint(std::move(expr2));
+
+  std::vector<F> public_trace = {};
+  math::RowMajorMatrix<F> main_trace(4, 2);
+  math::RowMajorMatrix<F> preprocessed_trace(4, 2);
+
+  // clang-format off
+  main_trace << F(5), F(0),
+                F(5), F(4),
+                F(5), F(4),
+                F(0), F(4);
+
+  preprocessed_trace << F(1), F(0),
+                        F(1), F(1),
+                        F(0), F(1),
+                        F(0), F(1);
+  // clang-format on
+
+  AirEvaluator<F> evaluator;
+  bool is_satisfied = constraint_system.IsSatisfied(
+      evaluator, public_trace, main_trace, &preprocessed_trace);
+  EXPECT_TRUE(is_satisfied);
+}
+
+}  // namespace tachyon::zk::air
--- a/tachyon/zk/air/constraint_system/variable.h
+++ b/tachyon/zk/air/constraint_system/variable.h
@@ -1,0 +1,101 @@
+#ifndef TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_VARIABLE_H_
+#define TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_VARIABLE_H_
+
+#include <string>
+
+#include "absl/strings/substitute.h"
+
+#include "tachyon/base/logging.h"
+#include "tachyon/export.h"
+
+namespace tachyon::zk::air {
+
+class TACHYON_EXPORT Variable {
+ public:
+  enum class Type {
+    kPreprocessed,
+    kMain,
+    kPublic,
+    kPermutation,
+    kChallenge,
+  };
+
+  constexpr static Variable Preprocessed(size_t row_index, size_t col_index) {
+    return Variable(Type::kPreprocessed, row_index, col_index);
+  }
+  constexpr static Variable Main(size_t row_index, size_t col_index) {
+    return Variable(Type::kMain, row_index, col_index);
+  }
+  constexpr static Variable Public(size_t row_index) {
+    return Variable(Type::kPublic, row_index, 0);
+  }
+  constexpr static Variable Permutation(size_t row_index,
+                                        size_t col_index = 0) {
+    return Variable(Type::kPermutation, row_index, col_index);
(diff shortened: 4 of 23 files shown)


tachyon/zk/air/expressions/evaluator.h
```diff
@@ -0,0 +1,100 @@
+#ifndef TACHYON_ZK_AIR_EXPRESSIONS_EVALUATOR_H_
+#define TACHYON_ZK_AIR_EXPRESSIONS_EVALUATOR_H_
+
+#include <vector>
+
+#include "tachyon/base/logging.h"
+#include "tachyon/math/matrix/matrix_types.h"
```
ashjeong, line 7: Unused I believe? If this is included due to `Eigen::Index`, perhaps `#include "third_party/eigen3/Eigen/Core"` would be better suited.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,140 @@
...
+#define TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
+
+#include <stddef.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "tachyon/math/matrix/matrix_types.h"
+#include "tachyon/math/matrix/prime_field_num_traits.h"
```
ashjeong, line 11: Replace with `#include "third_party/eigen3/Eigen/Core"`

tachyon/zk/air/constraint_system/constraint_system_unittest.cc
```diff
@@ -0,0 +1,64 @@
+#include "tachyon/zk/air/constraint_system/constraint_system.h"
+
+#include "gtest/gtest.h"
+
+#include "tachyon/math/finite_fields/test/finite_field_test.h"
+#include "tachyon/math/finite_fields/test/gf7.h"
+#include "tachyon/math/matrix/matrix_types.h"
+#include "tachyon/math/matrix/prime_field_num_traits.h"
```
ashjeong, line 8: Remove

batzor: Here, `matrix_types.h` is used for including `RowMajorMatrix`. I cleaned up on other headers

tachyon/zk/air/expressions/evaluator.h
```diff
@@ -0,0 +1,100 @@
...
+        NOTREACHED() << ExpressionTypeToString(input->type())
+                     << " expression is not allowed in AIR!";
+    }
+    NOTREACHED();
+    return F::Zero();
+  }
+};
+
+}  // namespace tachyon::zk::air
+#endif  // TACHYON_ZK_AIR_EXPRESSIONS_EVALUATOR_H_
```
dongchangYoo, line 100: it's minor
```suggestion
}  // namespace tachyon::zk::air

#endif  // TACHYON_ZK_AIR_EXPRESSIONS_EVALUATOR_H_
```

tachyon/zk/air/constraint_system/BUILD.bazel
```diff
@@ -1,7 +1,17 @@
-load("//bazel:tachyon_cc.bzl", "tachyon_cc_library")
+load("//bazel:tachyon_cc.bzl", "tachyon_cc_library", "tachyon_cc_unittest")
```
dongchangYoo, line 1: Please add a link the reference of this implementation to commit body.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+    for (size_t i = 0; i < num_public_values; ++i) {
+      public_values_.push_back(Variable::Public(i));
+    }
+  }
+
+  const std::vector<Variable>& GetPublicValues() const {
+    return public_values_;
+  }
```
dongchangYoo, line 48: In our convention, it seems more appropriate to name this function `public_values()`.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,140 @@
...
+
+  const std::vector<Variable>& GetPreprocessed() const { return preprocessed_; }
```
dongchangYoo, line 50: In our convention, it seems more appropriate to name this function `preprocessed()`.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,140 @@
...
+
+  const std::vector<Variable>& GetMain() const { return main_; }
```
dongchangYoo, line 52: In our convention, it seems more appropriate to name this function `main()`.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,140 @@
...
+    EnforceConstraint(ExpressionFactory<F>::Transition(std::move(constraint)));
+  }
+
+  void EnforceConstraint(std::unique_ptr<Expression<F>> constraint) {
+    constraints_.push_back(std::move(constraint));
+  }
+
+  template <typename MainDerived>
+  bool IsSatisfied(Evaluator<F>& evaluator, const std::vector<F>& public_values,
+                   std::reference_wrapper<const MainDerived> main) const {
```
dongchangYoo, line 72: I'm curious about the reason for intentionally using `std::reference_wrapper<const MainDerived>`, instead of `const MainDerived& main`

dongchangYoo (changes requested):

tachyon/zk/air/constraint_system/constraint_system.h
batzor: @dongchangYoo It was leftover from my attempt of removing the template and allowing all matrix types ;). But I finally figured it out and now it has no raw pointers, non-template and still zero-copy. 

Reference: https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html under "How to write generic, but non-templated function?"

batzor: Sorry it was leftover from my attempt from taking optional matrix as `std::optional<RowMajorMatrix<F>&>`. I cleaned it up and refactored evaluator to remove raw pointer usage.

tachyon/zk/air/expressions/evaluator.h
batzor: Now, evaluator takes `RowMajorMatrix<F>&` instead of `const F *`

tachyon/zk/air/constraint_system/constraint_system.h
dongchangYoo: If you need a `RowMajorMatrix<F>` in the `None` state, how about using a pointer?

chokobole:
In https://github.com/kroma-network/tachyon/pull/453/commits/f196ed41fc006f9d734df3db36dcc3b0b15a350d, "Closes" should be in commit body. e.g, See debd9c677ae51cd155c91a5128e56030189c3aeb

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+  }
+
+ private:
+  Eigen::Index preprocessed_width_ = 0;
+  Eigen::Index main_width_ = 2;
+  size_t num_public_values_ = 0;
+  std::vector<Variable> preprocessed_;
+  std::vector<Variable> main_;
+  std::vector<Variable> public_values_;
```
chokobole, line 115: Why they are needed?

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,121 @@
...
+
+  bool IsSatisfied(Evaluator<F>& evaluator, const std::vector<F>& public_values,
+                   const math::RowMajorMatrix<F>& preprocessed,
+                   const math::RowMajorMatrix<F>& main) const {
+    CHECK(public_values.size() == num_public_values_);
+    CHECK(main.cols() == main_width_);
+    CHECK(preprocessed.cols() == preprocessed_width_);
+    CHECK(preprocessed.rows() == main.rows());
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
```
chokobole, line 95: Why not `    for (Eigen::Index i = 0; i < main.rows(); ++i) {`?

tachyon/zk/air/expressions/evaluator.h
```diff
@@ -0,0 +1,112 @@
...
+      case ExpressionType::kFirstRow: {
+        if (current_row == 0) {
+          const FirstRowExpression<F>* first_row = input->ToFirstRow();
+          return Evaluate(first_row->expr(), current_row, num_rows,
+                          public_values, preprocessed_window, main_window);
+        }
+        return F::Zero();
+      }
+      case ExpressionType::kLastRow: {
+        if (current_row == num_rows - 2) {
```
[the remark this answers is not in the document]
batzor: In `ConstraintSystem::IsSatisfied()`, it only iterates until `num_rows - 2` while in Plonky3, it iterates until `num_rows - 1` but [the second row of the window points to first row of the trace](https://github.com/Plonky3/Plonky3/blob/7c5deb0eab7191a97f7bb088637c1a68b2e6eb68/uni-stark/src/check_constraints.rs#L19) in the final iteration. I found it kind of redundant since we can just use the last transition window $(row_{\text{numrows}-2}, row_{numrows-1})$ for enforcing `LastRowConstraint`.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,121 @@
...
+    CHECK(public_values.size() == num_public_values_);
+    CHECK(main.cols() == main_width_);
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
+      for (const auto& constraint : constraints_) {
+        if (!evaluator
+                 .Evaluate(constraint.get(), i, main.rows(), public_values,
+                           std::nullopt, main.middleRows(i, 2))
+                 .IsZero()) {
+          NOTREACHED() << i << constraint->ToString();
```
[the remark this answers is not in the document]
batzor: I added it for debugging which constraint failed. Will remove.

tachyon/zk/air/constraint_system/constraint_system.h
batzor: Indeed they are not needed if we manually construct `Variable`when building the constraints. This is taken from [Plonky3 implementation.](https://github.com/Plonky3/Plonky3/blob/7c5deb0eab7191a97f7bb088637c1a68b2e6eb68/uni-stark/src/symbolic_builder.rs#L76). Do you want me to remove?

tachyon/zk/air/expressions/evaluator.h
```diff
@@ -0,0 +1,112 @@
...
+#include "tachyon/zk/air/expressions/first_row_expression.h"
+#include "tachyon/zk/air/expressions/last_row_expression.h"
+#include "tachyon/zk/air/expressions/transition_expression.h"
+#include "tachyon/zk/air/expressions/variable_expression.h"
+#include "tachyon/zk/expressions/expression.h"
+
+namespace tachyon::zk::air {
+
+template <typename F>
+class Evaluator {
```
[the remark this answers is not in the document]
batzor: Hmm! So do you want me to have another function (ie. `SetWindowContext()`) to set those member variables before every call? Also, I would like to remind that [plonk evaluators](https://github.com/kroma-network/tachyon/blob/main/tachyon/zk/plonk/expressions/evaluator/simple_selector_extractor.h) are not derived from `Evaluator` class anymore.

tachyon/zk/air/constraint_system/variable.h
```diff
@@ -0,0 +1,77 @@
+#ifndef TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_VARIABLE_H_
+#define TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_VARIABLE_H_
+
+#include <stddef.h>
+
+#include <vector>
+
+#include "tachyon/export.h"
```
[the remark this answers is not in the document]
batzor: I'm not sure why `logging.h` is needed here.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,121 @@
...
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  bool IsSatisfied(Evaluator<F>& evaluator, const std::vector<F>& public_values,
+                   const math::RowMajorMatrix<F>& preprocessed,
+                   const math::RowMajorMatrix<F>& main) const {
```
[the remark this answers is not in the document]
batzor: I wanted to avoid raw pointer usage if possible. Should I make the change?

tachyon/zk/air/constraint_system/constraint_system.h
batzor: When `i == main.rows() - 1`, the transition window will be `(last_row, first_row)`. This case is used in Plonky3 for `LastRowConstraint` but we utilize the previous window for checking the `LastRowConstraint`.

tachyon/zk/air/constraint_system/constraint_system.h
chokobole: Yes we don't need to follow all if the things are unneeded.

tachyon/zk/air/constraint_system/variable.h
```diff
@@ -0,0 +1,77 @@
...
+      case Type::kPreprocessed:
+      case Type::kMain:
+      case Type::kPublic:
+        return 1;
+      case Type::kPermutation:
+      case Type::kChallenge:
+        return 0;
+      default:
+        NOTREACHED();
+        return 0;
```
chokobole, line 54: Please do not use `default` case, then compiler doesn't catch our mistake.

tachyon/zk/air/constraint_system/variable.h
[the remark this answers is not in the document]
chokobole: This is because of `NOTREACHED()`!

tachyon/zk/air/expressions/evaluator.h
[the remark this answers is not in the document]
chokobole: Oh I got it. Still reducing the number of the function arguments helps readability.

tachyon/zk/air/constraint_system/constraint_system.h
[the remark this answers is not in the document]
chokobole: When reading https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs, pointer is used to indicate that the function argument is optional.

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

tachyon/zk/air/expressions/air_evaluator.h
```diff
@@ -0,0 +1,117 @@
...
+      case ExpressionType::kAdvice:
+      case ExpressionType::kChallenge:
+      case ExpressionType::kFixed:
+      case ExpressionType::kInstance:
+      case ExpressionType::kSelector:
+        NOTREACHED() << ExpressionTypeToString(input->type())
+                     << " expression is not allowed in AIR!";
+        return F::Zero();
+    }
+  }
```
chokobole:
```suggestion
   }
   NOTREACHED();
   return F::Zero();
  }
```

tachyon/zk/air/expressions/air_evaluator.h
```diff
@@ -0,0 +1,117 @@
...
+                            const math::RowMajorMatrix<F>* preprocessed_window,
+                            const math::RowMajorMatrix<F>* main_window) {
+    current_row_ = current_row;
+    preprocessed_window_ = preprocessed_window;
+    main_window_ = main_window;
+  }
+
+ private:
+  Eigen::Index num_rows_;
+  const std::vector<F>* public_values_;
```
chokobole, line 110:
```suggestion
  absl::Span<const F> public_values_;
```

tachyon/zk/air/expressions/air_evaluator.h
```diff
@@ -0,0 +1,117 @@
...
+      case ExpressionType::kSelector:
+        NOTREACHED() << ExpressionTypeToString(input->type())
+                     << " expression is not allowed in AIR!";
+        return F::Zero();
+    }
+  }
+
+  void set_public_values(const std::vector<F>& public_values) {
+    public_values_ = &public_values;
+  }
```
chokobole, line 96:
```suggestion
  void set_public_values(absl::Span<const F> public_values) {
    public_values_ = public_values;
  }
```

tachyon/zk/air/expressions/air_evaluator.h
```diff
@@ -0,0 +1,117 @@
...
+    main_window_ = main_window;
+  }
+
+ private:
+  Eigen::Index num_rows_;
+  const std::vector<F>* public_values_;
+
+  Eigen::Index current_row_;
+  const math::RowMajorMatrix<F>* preprocessed_window_;
+  const math::RowMajorMatrix<F>* main_window_;
```
chokobole, line 114:
```suggestion
  // not owned
  const math::RowMajorMatrix<F>* preprocessed_window_ = nullptr;
  const math::RowMajorMatrix<F>* main_window_ = nullptr;
```

tachyon/zk/air/expressions/air_evaluator.h
```diff
@@ -0,0 +1,117 @@
...
+  void SetCurrentWindowData(Eigen::Index current_row,
+                            const math::RowMajorMatrix<F>* preprocessed_window,
+                            const math::RowMajorMatrix<F>* main_window) {
+    current_row_ = current_row;
+    preprocessed_window_ = preprocessed_window;
+    main_window_ = main_window;
+  }
+
+ private:
+  Eigen::Index num_rows_;
```
chokobole, line 109:
```suggestion
  Eigen::Index num_rows_ = 0;
```

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,95 @@
...
+                   const std::vector<F>& public_values,
+                   std::unique_ptr<const math::RowMajorMatrix<F>> preprocessed,
+                   const math::RowMajorMatrix<F>& main) const {
+    CHECK_EQ(public_values.size(), num_public_values_);
+    CHECK_EQ(main.cols(), main_width_);
+    CHECK(preprocessed != nullptr || preprocessed_width_ == 0);
+    if (preprocessed) {
+      CHECK_EQ(preprocessed->cols(), preprocessed_width_);
+      CHECK_EQ(preprocessed->rows(), main.rows());
+    }
```
chokobole, line 54:
```suggestion
    if (preprocessed) {
      CHECK_EQ(preprocessed->cols(), preprocessed_width_);
      CHECK_EQ(preprocessed->rows(), main.rows());
    } else {
      CHECK_EQ(processed_width_, 0);
    }
```

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,95 @@
...
+        }
+      }
+    }
+    return true;
+  }
+
+ private:
+  Eigen::Index preprocessed_width_ = 0;
+  Eigen::Index main_width_ = 0;
+  size_t num_public_values_ = 0;
```
chokobole, line 89:
```suggestion
  const Eigen::Index preprocessed_width_ = 0;
  const Eigen::Index main_width_ = 0;
  const size_t num_public_values_ = 0;
```

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,95 @@
+#ifndef TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
+#define TACHYON_ZK_AIR_CONSTRAINT_SYSTEM_CONSTRAINT_SYSTEM_H_
+
+#include <stddef.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "third_party/eigen3/Eigen/Core"
```
chokobole, line 10: This is minor, though. This seems redundant, cause `matrix_types.h` already includes this.

tachyon/zk/air/constraint_system/BUILD.bazel
```diff
@@ -1,7 +1,18 @@
...
 package(default_visibility = ["//visibility:public"])
 
+tachyon_cc_library(
+    name = "constraint_system",
+    hdrs = ["constraint_system.h"],
+    deps = [
+        ":variable",
+        "//tachyon/math/matrix:matrix_types",
+        "//tachyon/zk/air/expressions:air_evaluator",
+        "@eigen_archive//:eigen3",
```
chokobole, line 12: This is minor, though. This seems redundant, cause `//tachyon/math/matrix:matrix_types` already depends on this.

tachyon/zk/air/constraint_system/BUILD.bazel
```diff
@@ -11,3 +22,16 @@ tachyon_cc_library(
...
+
+tachyon_cc_unittest(
+    name = "air_unittests",
+    srcs = ["constraint_system_unittest.cc"],
+    deps = [
+        ":constraint_system",
+        "//tachyon/math/finite_fields/test:finite_field_test",
+        "//tachyon/math/finite_fields/test:gf7",
+        "//tachyon/math/matrix:matrix_types",
+        "//tachyon/zk/air/expressions:air_evaluator",
```
chokobole, line 34: These seems redundant since `constaint_system` already depends on them.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  void EnforceConstraint(std::unique_ptr<Expression<F>> constraint) {
+    constraints_.push_back(std::move(constraint));
+  }
+
+  bool IsSatisfied(AirEvaluator<F>& evaluator,
+                   const std::vector<F>& public_values,
+                   std::unique_ptr<const math::RowMajorMatrix<F>> preprocessed,
+                   const math::RowMajorMatrix<F>& main) const {
```
chokobole, line 47: Please do not transfer ownership since this function doesn't own the matrix in the end. Just pass it as a pointer. Also, please place the optional argument at the end.

```suggestion
                   const math::RowMajorMatrix<F>& main
                   const math::RowMajorMatrix<F>* preprocessed = nullptr) const {
```

chokobole: Also, I found that your code prioritizes `preprocessed` over `main`, but considering that `preprocessed` is optional, the order should be the other way around.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,95 @@
...
+        }
+      } else {
+        evaluator.SetCurrentWindowData(i, nullptr, &main_window);
+
+        for (const std::unique_ptr<Expression<F>>& constraint : constraints_) {
+          if (!evaluator.Evaluate(constraint.get()).IsZero()) {
+            return false;
+          }
+        }
+      }
```
chokobole, line 81:
```suggestion
      const math::RowMajorMatrix<F>& main_window = main.middleRows(i, 2);
      const math::RowMajorMatrix<F>& prep_window =
          preprocessed? preprocessed->middleRows(i, 2): nullptr;

      evaluator.SetCurrentWindowData(i, &prep_window, &main_window);

      for (const std::unique_ptr<Expression<F>>& constraint : constraints_) {
        if (!evaluator.Evaluate(constraint.get()).IsZero()) {
          return false;
        }
      }
```

tachyon/zk/air/constraint_system/constraint_system.h
batzor: Hmm I try to keep transitive includes to minimum following https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

But I agree with you that sometimes it's too verbose. Should I remove?

tachyon/zk/air/constraint_system/constraint_system.h
batzor: I don't think this would compile since you can't assign nullptr to reference.

```
const math::RowMajorMatrix<F>* prep_window =
	          preprocessed ? &preprocessed->middleRows(i, 2): nullptr;
```
	          
would not work as well since object will be out of scope.

chokobole: Oh, you're right. What I tried to say was that if possible, please reduce code duplication!

batzor: I think it's hard since `Evaluate` must be called within the `if (preprocessed) {}` scope or `prep_window` will go out of scope and result UB.

chokobole: Since it's minor, i'll resolve.

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,94 @@
...
+
+    evaluator.set_public_values(public_values);
+    evaluator.set_num_rows(main.rows());
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
+      const math::RowMajorMatrix<F>& main_window = main.middleRows(i, 2);
+
+      if (preprocessed) {
+        const math::RowMajorMatrix<F>& prep_window =
+            preprocessed->middleRows(i, 2);
```
chokobole, line 63: If you take this as `Eigen::Matrix`, this will cause a copy. Please modify the type to `Eigen::Block<>`.

```suggestion
        Eigen::Block<math::RowMajorMatrix<F>> prep_window =
            preprocessed->middleRows(i, 2);
```

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,94 @@
...
+      CHECK_EQ(preprocessed->rows(), main.rows());
+    } else {
+      CHECK_EQ(preprocessed_width_, 0);
+    }
+
+    evaluator.set_public_values(public_values);
+    evaluator.set_num_rows(main.rows());
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
+      const math::RowMajorMatrix<F>& main_window = main.middleRows(i, 2);
```
chokobole, line 59: If you take this as `Eigen::Matrix`, this will cause a copy. Please modify the type to `Eigen::Block<>`.
If you take this as `Eigen::Matrix`, this will cause a copy. Please modify the type to `Eigen::Block<>`.

```suggestion
      Eigen::Block<math::RowMajorMatrix<F>> main_window = main.middleRows(i, 2);
```

tachyon/zk/air/constraint_system/constraint_system.h
```diff
@@ -0,0 +1,88 @@
...
+
+    for (Eigen::Index i = 0; i < main.rows() - 1; ++i) {
+      const ConstMatrixBlock main_window(main, i, 0, 2, main_width_);
+      const ConstMatrixBlock* prep_window = nullptr;
+
+      if (preprocessed) {
+        prep_window =
+            new ConstMatrixBlock(*preprocessed, i, 0, 2, preprocessed_width_);
+      }
+      evaluator.SetCurrentWindowData(i, &main_window, prep_window);
```
chokobole, line 69:
```suggestion
      if (preprocessed) {
        ConstMatrixBlock prep_window(*preprocessed, i, 0, 2, preprocessed_width_);      
        evaluator.SetCurrentWindowData(i, &main_window, prep_window);
      } else {
        evaluator.SetCurrentWindowData(i, &main_window, nullptr);
      }
```

batzor: `prep_window` will be destructed outside of the `if` so it will cause UB on the `Evaluate` loop.

chokobole: Oh I see. then could we allocate heap just once and reuse it?

chokobole (approved):

PR #456, opened 2024-07-02
perf: optimize fft

GiDeokKim:
# Description

This PR optimizes the `DoFFT` function by decreasing unnecessary swap operations as much as possible.

### DIFF
--- a/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
+++ b/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
@@ -174,12 +174,10 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
     return absl::WrapUnique(new Radix2EvaluationDomain(*this));
   }
 
-  // UnivariateEvaluationDomain methods
   CONSTEXPR_IF_NOT_OPENMP void DoFFT(Evals& evals) const override {
     DegreeAwareFFTInPlace(evals);
   }
 
-  // UnivariateEvaluationDomain methods
   CONSTEXPR_IF_NOT_OPENMP void DoIFFT(DensePoly& poly) const override {
     poly.coefficients_.coefficients_.resize(this->size_, F::Zero());
     InOrderIFFTInPlace(poly);
--- a/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
+++ b/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
@@ -199,7 +199,7 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
     uint32_t log_d = base::bits::SafeLog2Ceiling(num_coeffs);
     // When the polynomial is of size k * |coset|, for k < 2ⁱ, the first i
     // iterations of Cooley-Tukey are easily predictable. This is because they
-    // will be combining g(w²) + wh(w²), but g or h will always refer to a
+    // will be combining g(ω²) + ω * h(ω²), but g or h will always refer to a
     // coefficient that is 0. Therefore those first i rounds have the effect
     // of copying the evaluations into more locations, so we handle this in
     // initialization, and reduce the number of loops that are performing
--- a/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
+++ b/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
@@ -180,7 +180,6 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
         this->size_) {
       DegreeAwareFFTInPlace(evals);
     } else {
-      evals.evaluations_.resize(this->size_, F::Zero());
       InOrderFFTInPlace(evals);
     }
   }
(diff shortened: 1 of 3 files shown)


chokobole:
- In https://github.com/kroma-network/tachyon/pull/456/commits/e06d337c7374fd862c47f9fd45e515873726564c and https://github.com/kroma-network/tachyon/pull/456/commits/e909250703f3ecd986589b9fe2f45565da42697c, please elaborate the motivation in commit body.
- In https://github.com/kroma-network/tachyon/pull/456/commits/5e24e98efe93d24865956171a63f3318774da4f6, please change the commit message to **chore(math)**: change omega character
- In https://github.com/kroma-network/tachyon/pull/456/commits/7ee26cb2c352f5e65cc8e0ccc89479e3b931ec11, please change the commit message to **chore(math)**: remove redundant comments
- please squash https://github.com/kroma-network/tachyon/pull/456/commits/f9788d0e52a2628361adff8d9e3d628e40cf28c4 to https://github.com/kroma-network/tachyon/pull/456/commits/e909250703f3ecd986589b9fe2f45565da42697c

dongchangYoo (approved):
LGTM

chokobole (approved):
LGTM

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -250,9 +249,10 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
   }
 
   constexpr void FFTHelperInPlace(Evals& evals) const {
-    uint32_t log_len = static_cast<uint32_t>(base::bits::Log2Ceiling(
-        static_cast<uint32_t>(evals.evaluations_.size())));
-    this->SwapElements(evals, evals.evaluations_.size() - 1, log_len);
+    size_t num_coeffs = vals.evaluations_.size();
+    uint32_t log_n = this->log_size_of_group_;
+    evals.evaluations_.resize(this->size_, F::Zero());
```
TomTaehoonKim: I'm not sure if we are optimizing the use of swapElements() if we are filling `F::Zero()`s.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -214,26 +209,23 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
     size_t duplicity_of_initials = size_t{1} << (log_n - log_d);
     evals.evaluations_.resize(n, F::Zero());
     this->SwapElements(evals, num_coeffs, log_n);
-    if (duplicity_of_initials > 1) {
+    size_t start_gap = 1;
+    // NOTE(GideokKim): |base::ParallelizeByChunkSize(container, chunk_size,
```
TomTaehoonKim, line 211:
```suggestion
    // NOTE(GiDeokKim): |base::ParallelizeByChunkSize(container, chunk_size,
```

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -214,26 +209,23 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
...
+    // callback)| is not called when |duplicity_of_initials| is 2 or 3. Since we
+    // have not benchmarked this case, it is called only when it is 4 or higher.
```
ashjeong, line 213: I think simplifying the full statement to this would be better; however, I'm not sure if this comment is needed, since users can go see that `kDegreeAwareFFTThresholdFactor` is 4.
```suggestion
    // NOTE(GideokKim): |base::ParallelizeByChunkSize(container, chunk_size,
    // callback)| is only called when |duplicity_of_initials| is 4 or higher.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
GiDeokKim: Looking at the arguments of `SwapElements()`, the second argument is a value derived from **_the size of the polynomial_**, and the third argument is a value derived from **_the domain size_**. The swap operation is ultimately performed within **_the domain size_**, but it is true that it is optimized because it is operated only as large as **_the size of the polynomial_** given by the second argument. 

If we do not use the second argument as `num_coeffs`, which is the value before `resize()`, a situation where `F:Zero()` is swapped may occur. Also, if we run the `SwapElements()` function without filling it with `F:Zero()`, the swap operation goes outside of size (since **_the size of the polynomial_** <= **_the domain size_**).

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
GiDeokKim: I think the same thing. If we are just going to write that it's called only when `duplicity_of_initials` is 4 or higher, I don't think this comment is needed.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
GiDeokKim: I deleted this comment.

ashjeong (approved):
LGTM

TomTaehoonKim (approved):
LGTM

PR #463, opened 2024-07-07
perf: use msm gpu when generating groth16 proof

chokobole:
# Description

This PR uses msm gpu when generating groth16 proofs.

The benchmarking result based on circom proof generation received from Orbiter Finance Team are as follows:

```
# Gpu
Run 1, Time Taken: {Parse ZKey: 29.472 s, Parse Wtns: 0.505613 s, Prove: 14.5769 s, Verify: 0.005145 s} 
# Cpu
Run 1, Time Taken: {Parse ZKey: 29.4742 s, Parse Wtns: 0.498701 s, Prove: 32.4379 s, Verify: 0.001405 s}
```


TomTaehoonKim (approved):
LGTM. Please update the "Speed" section in circom README after GPU benchmark.

tachyon/math/elliptic_curves/msm/variable_base_msm_gpu_unittest.cc
```diff
@@ -7,7 +7,6 @@
 #include "tachyon/device/gpu/gpu_enums.h"
 #include "tachyon/device/gpu/scoped_mem_pool.h"
 #include "tachyon/device/gpu/scoped_stream.h"
-#include "tachyon/math/elliptic_curves/bn/bn254/g1.h"
```
ashjeong, line 10: Please also delete this dependency in the build file

tachyon/zk/r1cs/groth16/prove.h
```diff
@@ -46,11 +52,39 @@ Proof<Curve> CreateProofWithAssignment(const ProvingKey<Curve>& pk, const F& r,
...
+  device::gpu::ScopedStream stream = device::gpu::CreateStream();
+
+  math::VariableBaseMSMGpu<G1AffinePoint> msm(mem_pool.get(), stream.get());
+
+  device::gpu::ScopedMemPool mem_pool2 = device::gpu::CreateMemPool(&props);
+
+  error = gpuMemPoolSetAttribute(
+      mem_pool2.get(), gpuMemPoolAttrReleaseThreshold, &mem_pool_threshold);
+  CHECK_EQ(error, gpuSuccess);
+  device::gpu::ScopedStream stream2 = device::gpu::CreateStream();
```
ashjeong, line 78: Can we reuse `mem_pool` and `stream` as well?

ashjeong (changes requested):
https://github.com/kroma-network/tachyon/pull/463/commits/49d3787bf7e8f1cd99f7062b59d5f19c31d89fe7
Remove the “to” in the commit title

tachyon/zk/r1cs/groth16/prove.h
chokobole: Maybe for `mem_pool`, but for `stream`, I think it's better to separate as many as possible. See https://developer.download.nvidia.com/CUDA/training/StreamsAndConcurrencyWebinar.pdf I want to listen to @GideokKim 's opinion.

tachyon/zk/r1cs/groth16/prove.h
```diff
@@ -50,6 +50,7 @@ Proof<Curve> CreateProofWithAssignment(const ProvingKey<Curve>& pk, const F& r,
   using G2Bucket = typename math::VariableBaseMSM<G2AffinePoint>::Bucket;
 
   math::VariableBaseMSM<G1AffinePoint> msm;
+  math::VariableBaseMSM<G2AffinePoint> msm2;
```
batzor, line 86: How about renaming it `msm_g1`, `msm_g2`?

chokobole:
> LGTM. Please update the "Speed" section in circom README after GPU benchmark.

Since MSM Gpu doesn't improve the performance on `sha256_512.circom`, we need to find additional circom file to benchmark to show performance gain. Maybe https://github.com/kroma-network/tachyon/tree/benchmark/plonky2-circom is one candidate, the problem is the size of the files are too large..

batzor:
LGTM

batzor (approved):

dongchangYoo (approved):
LGTM

tachyon/zk/r1cs/constraint_system/quadratic_arithmetic_program.h
```diff
@@ -26,19 +26,15 @@ namespace tachyon::zk::r1cs {
...
-  return std::accumulate(sums.begin(), sums.end(), F::Zero(), std::plus<>());
+  F sum;
+  for (const Cell<F>& cell : cells) {
+    if (cell.coefficient.IsOne()) {
+      sum += assignments[cell.index];
+    } else {
+      sum += assignments[cell.index] * cell.coefficient;
+    }
+  }
+  return sum;
```
ashjeong, line 37: Could you make this into a separate commit?

chokobole: Actually, this is part of the commit, since this used parallel loop within in an outer parallel loop!

tachyon/zk/r1cs/groth16/prove.h
GideokKim: ### It is best to set up multiple streams.

If two kernel functions exist in the same stream, the kernel functions are necessarily executed sequentially. This means that the second kernel function will not be executed unless the execution of the previous kernel function is completed.

### In the case of memory pools, it is better not to be identical.

When using the same memory pool in different streams, record events using `cudaEventRecord(event, stream);` for one stream and `cudaStreamWaitEvent(streamB, event, 0);` for the other stream. It is recommended to use to wait for an event. I think it's better to let the CUDA driver do the optimization on its own. The CUDA driver can reuse memory opportunistically in the absence of explicit dependencies specified by the application.

GideokKim (approved):
LGTM

ashjeong (approved):
LGTM!

PR #467, opened 2024-07-11
fix!(halo2): remove constraint system serialization from `vk.write()`

chokobole:
# Description

Scroll-prover (de)serializes `VerifyingKey` and constructs `constraint_system` from `configure()` instead of the buffer. Therefore, `constraint_system` should be removed from the serialization process. However, since Tachyon depends on `constraint_system` serialization, an additional function `write_including_cs()` has been added to handle this requirement.

BREAKING CHANGE: `pk.write()` should be replaced with `pk.write_including_cs()` to serialize proving key.


chokobole:
It seems that the crash in the scroll-prover was caused by an issue in the serialization and deserialization of the VerifyingKey. Specifically, the [`read()` function in halo2_proofs does not read `cs` but instead gets this value by configuring it](https://github.com/kroma-network/halo2/blob/89e632234dd3ebee4c8fcf0054908d1fc0d0cbf4/halo2_proofs/src/plonk.rs#L94-L140). However, in the code, [`cs` was written between `fixed_commitments` and `permutation`](https://github.com/kroma-network/halo2/blob/89e632234dd3ebee4c8fcf0054908d1fc0d0cbf4/halo2_proofs/src/plonk.rs#L73-L92), which led to the deserialization error. This issue has been resolved in a recent commit in the halo2 repository.

```shell
thread 'main' panicked at /home/chokobole/.cargo/git/checkouts/zkevm-circuits-192f22d2e9118534/c38763c/prover/src/io.rs:93:10:
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Invalid point encoding in proof" }
```

See also https://github.com/scroll-tech/zkevm-circuits/blob/ede39de59ce390ca6d3b1f0568e8661e4868f4fc/prover/src/io.rs#L97-L106.

TomTaehoonKim (approved):
LGTM

GideokKim (approved):
LGTM

batzor (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #473, opened 2024-07-18
feat: enable pedersen gpu

chokobole:
# Description

This PR enables pedersen commitment using GPU.

### DIFF
--- a/tachyon/crypto/commitments/pedersen/pedersen.h
+++ b/tachyon/crypto/commitments/pedersen/pedersen.h
@@ -8,6 +8,9 @@
 
 #include <stddef.h>
 
+#include <algorithm>
+#include <limits>
+#include <memory>
 #include <sstream>
 #include <string>
 #include <utility>
@@ -19,6 +22,12 @@
 #include "tachyon/crypto/commitments/vector_commitment_scheme.h"
 #include "tachyon/math/elliptic_curves/msm/variable_base_msm.h"
 #include "tachyon/math/elliptic_curves/point_conversions.h"
+
+#if TACHYON_CUDA
+#include "tachyon/device/gpu/scoped_mem_pool.h"
+#include "tachyon/device/gpu/scoped_stream.h"
+#include "tachyon/math/elliptic_curves/msm/variable_base_msm_gpu.h"
+#endif
 
 namespace tachyon {
 namespace crypto {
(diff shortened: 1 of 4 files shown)


tachyon/crypto/commitments/pedersen/pedersen_unittest.cc
```diff
@@ -29,15 +29,24 @@ TEST_F(PedersenTest, CommitPedersen) {
       base::CreateVector(kMaxSize, []() { return math::bn254::Fr::Random(); });
 
   math::bn254::Fr r = math::bn254::Fr::Random();
-  math::bn254::G1AffinePoint commitment;
-  ASSERT_TRUE(vcs.Commit(v, r, &commitment));
+  math::bn254::G1AffinePoint cpu_commitment;
+  ASSERT_TRUE(vcs.Commit(v, r, &cpu_commitment));
```
TomTaehoonKim, line 33: Do you think we need to test both cpu and gpu for cuda tests?

chokobole: I think so, because both are different code.

tachyon/crypto/commitments/pedersen/pedersen.h
```diff
@@ -136,7 +155,6 @@ struct VectorCommitmentSchemeTraits<Pedersen<Point, MaxSize, _Commitment>> {
 }  // namespace crypto
 
 namespace base {
-
```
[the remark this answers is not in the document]
chokobole: It was my mistake, thanks for pointing out it!

tachyon/crypto/commitments/pedersen/pedersen_unittest.cc
```diff
@@ -28,14 +29,15 @@ TEST_F(PedersenTest, CommitPedersen) {
...
-  math::VariableBaseMSM<math::bn254::G1JacobianPoint> msm;
-  math::bn254::G1JacobianPoint msm_result;
-  ASSERT_TRUE(msm.Run(vcs.generators(), v, &msm_result));
+  math::VariableBaseMSM<math::bn254::G1AffinePoint> msm;
+  math::bn254::G1PointXYZZ msm_result_tmp;
+  ASSERT_TRUE(msm.Run(vcs.generators(), v, &msm_result_tmp));
+  math::bn254::G1AffinePoint msm_result = msm_result_tmp.ToAffine();
 
-  EXPECT_EQ(commitment, msm_result + r * vcs.h());
+  EXPECT_EQ(commitment, (msm_result + r * vcs.h()).ToAffine());
```
ashjeong: Perhaps to name the variables more descriptively, `msm_result_xyzz` instead of `msm_result_tmp` and `men_result_affine` instead of `msm_result` would be better.  I also wanted to ask why `(msm_result + r * vcs.h()).ToAffine()` needs to be converted to affine if `msm_result` is already affine?

tachyon/crypto/commitments/pedersen/pedersen_unittest.cc
```diff
@@ -57,17 +59,19 @@ TEST_F(PedersenTest, BatchCommitPedersen) {
...
-  msm_results.resize(num_vectors);
+  math::VariableBaseMSM<math::bn254::G1AffinePoint> msm;
+  std::vector<math::bn254::G1PointXYZZ> msm_results_tmp(num_vectors);
   for (size_t i = 0; i < num_vectors; ++i) {
-    ASSERT_TRUE(msm.Run(vcs.generators(), v_vec[i], &msm_results[i]));
+    ASSERT_TRUE(msm.Run(vcs.generators(), v_vec[i], &msm_results_tmp[i]));
   }
+  std::vector<math::bn254::G1AffinePoint> msm_results(num_vectors);
+  ASSERT_TRUE(
+      math::bn254::G1PointXYZZ::BatchNormalize(msm_results_tmp, &msm_results));
```
ashjeong: ^^ Same naming concern here

batzor (approved):
LGTM

tachyon/crypto/commitments/pedersen/pedersen_unittest.cc
TomTaehoonKim: Oh I was trying to say I think we only need cpu test when cuda is not enabled and we only need gpu test when cuda is enabled.

chokobole: Hmm, but `KZG` already tests like that, and since it doesn't take that long, i think it's okay!

tachyon/crypto/commitments/pedersen/pedersen_unittest.cc
chokobole: Since `Affine` + `Affine` results in `Jacobian`, See the return type of `AffinePoint<>::Add(const AffinePoint&)` in `tachyon/math/elliptic_curves/short_weierstrass/affine_point.h`

TomTaehoonKim (approved):
LGTM

ashjeong (approved):
LGTM

dongchangYoo (approved):
LGTM

GideokKim (approved):
LGTM

PR #484, opened 2024-07-23
test: optimize benchmark test

chokobole:
# Description

In benchmarks, data is typically created using base::CreateVector(). However, since benchmarks generate a large amount of data, running this process in parallel is beneficial. This PR implements parallel data creation for benchmarks.

### DIFF
--- a/tachyon/base/containers/container_util.h
+++ b/tachyon/base/containers/container_util.h
@@ -12,6 +12,7 @@
 #include "tachyon/base/functional/functor_traits.h"
 #include "tachyon/base/logging.h"
 #include "tachyon/base/numerics/checked_math.h"
+#include "tachyon/base/openmp_util.h"
 #include "tachyon/base/random.h"
 
 namespace tachyon::base {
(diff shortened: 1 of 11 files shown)


TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

ashjeong (approved):
LGTM

tachyon/base/containers/container_util.h
```diff
@@ -47,6 +48,20 @@ std::vector<ReturnType> CreateVector(size_t size, Generator&& generator) {
...
+          typename FunctorTraits = internal::MakeFunctorTraits<Generator>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 0>* = nullptr>
+std::vector<ReturnType> CreateVectorParallel(size_t size,
+                                             Generator&& generator) {
+  std::vector<ReturnType> ret(size);
+  OPENMP_PARALLEL_FOR(size_t i = 0; i < size; ++i) { ret[i] = generator(); }
```
batzor, line 61: So `CreateVectorParallel` is only usable when `Generator` is thread-safe right? Maybe this should be added as comment.

tachyon/math/elliptic_curves/test/random.h
```diff
@@ -15,12 +15,17 @@ std::vector<Point> CreatePseudoRandomPoints(size_t size) {
...
-    return ret;
+  // TODO(chokobole): make |ConvertPoints()| use |BatchInverse()| internally,
+  // and use |ConvertPoints()|.
+  std::vector<Point> ret(size);
+  base::Parallelize(ret, [](absl::Span<Point> chunk) {
+    Point r = Point::Random();
+    for (Point& p : chunk) {
+      p = ConvertPoint<Point>(r);
+      r = ConvertPoint<Point>(r.Double());
+    }
```
batzor, line 26: Can't we use `CreateVectorParallel` here?

chokobole: I think that's slower since `Point::Random()` is expensive operation, what can be more optimized here is if we somehow know that `Point` is affine, then we can call `BatchNormalize()` instead of `ConvertPoint()`

batzor (approved):
LGTM

PR #486, opened 2024-07-24
feat: use pmr vector for custom allocation

batzor:
# Description

This PR does following:
- Implement utility functions for `std::pmr::vector`
- Switch `std::vector` to  `std::pmr::vector` in `UnivariateEvaluations` and `UnivariateDenseCoefficients`.

[Here is a quick benchmark of `std::pmr::vector` and `std::vector` ](https://quick-bench.com/q/EMkmBbUi0f1oISx1EIn1MY_LkcA)for simple char vector initialization. Following this, I am expecting ~10% improvement in large vector initialization and resize.

### DIFF
--- a/tachyon/base/buffer/copyable.h
+++ b/tachyon/base/buffer/copyable.h
@@ -2,6 +2,7 @@
 #define TACHYON_BASE_BUFFER_COPYABLE_H_
 
 #include <array>
+#include <memory_resource>
 #include <numeric>
 #include <string>
 #include <vector>
@@ -133,10 +134,10 @@ class Copyable<T[N]> {
   }
 };
 
-template <typename T>
-class Copyable<std::vector<T>> {
+template <typename V, typename T = typename V::value_type>
+class CopyableVectorImpl {
  public:
-  static bool WriteTo(const std::vector<T>& values, Buffer* buffer) {
+  static bool WriteTo(const V& values, Buffer* buffer) {
     if (!buffer->Write(values.size())) return false;
     for (const T& value : values) {
       if (!buffer->Write(value)) return false;
@@ -144,7 +145,7 @@ class Copyable<std::vector<T>> {
     return true;
   }
 
-  static bool ReadFrom(const ReadOnlyBuffer& buffer, std::vector<T>* values) {
+  static bool ReadFrom(const ReadOnlyBuffer& buffer, V* values) {
     size_t size;
     if (!buffer.Read(&size)) return false;
     values->resize(size);
@@ -154,13 +155,19 @@ class Copyable<std::vector<T>> {
     return true;
   }
 
-  static size_t EstimateSize(const std::vector<T>& values) {
+  static size_t EstimateSize(const V& values) {
     return std::accumulate(values.begin(), values.end(), sizeof(size_t),
                            [](size_t total, const T& value) {
                              return total + base::EstimateSize(value);
                            });
   }
 };
+
+template <typename T>
+class Copyable<std::vector<T>> : public CopyableVectorImpl<std::vector<T>> {};
+template <typename T>
+class Copyable<std::pmr::vector<T>>
+    : public CopyableVectorImpl<std::pmr::vector<T>> {};
 
 template <typename T, size_t N>
 class Copyable<std::array<T, N>> {
--- a/tachyon/base/containers/container_util.h
+++ b/tachyon/base/containers/container_util.h
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <iterator>
+#include <memory_resource>
 #include <optional>
 #include <utility>
 #include <vector>
@@ -97,6 +98,42 @@ std::vector<ReturnType> CreateVectorParallel(size_t size,
   return ret;
 }
 
+template <typename Generator,
+          typename FunctorTraits = internal::MakeFunctorTraits<Generator>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 0>* = nullptr>
+std::pmr::vector<ReturnType> CreatePmrVector(size_t size,
+                                             Generator&& generator) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(size);
+  std::generate_n(std::back_inserter(ret), size,
+                  std::forward<Generator>(generator));
+  return ret;
+}
+
+template <typename Generator,
+          typename FunctorTraits = internal::MakeFunctorTraits<Generator>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 1>* = nullptr>
+std::pmr::vector<ReturnType> CreatePmrVector(size_t size,
+                                             Generator&& generator) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(size);
+  size_t idx = 0;
+  std::generate_n(
+      std::back_inserter(ret), size,
+      [&idx, generator = std::forward<Generator>(generator)]() mutable {
+        return generator(idx++);
+      });
+  return ret;
+}
+
 template <typename Iterator, typename UnaryOp,
           typename FunctorTraits = internal::MakeFunctorTraits<UnaryOp>,
           typename ReturnType = typename FunctorTraits::ReturnType,
@@ -134,6 +171,47 @@ template <typename Container, typename UnaryOp>
 auto Map(Container&& container, UnaryOp&& op) {
   return Map(std::begin(container), std::end(container),
              std::forward<UnaryOp>(op));
+}
+
+template <typename Iterator, typename UnaryOp,
+          typename FunctorTraits = internal::MakeFunctorTraits<UnaryOp>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 1>* = nullptr>
+std::pmr::vector<ReturnType> PmrMap(Iterator begin, Iterator end,
+                                    UnaryOp&& op) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(std::distance(begin, end));
+  std::transform(begin, end, std::back_inserter(ret),
+                 std::forward<UnaryOp>(op));
+  return ret;
+}
+
+template <typename Iterator, typename UnaryOp,
+          typename FunctorTraits = internal::MakeFunctorTraits<UnaryOp>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 2>* = nullptr>
+std::pmr::vector<ReturnType> PmrMap(Iterator begin, Iterator end,
+                                    UnaryOp&& op) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(std::distance(begin, end));
+  size_t idx = 0;
+  std::transform(begin, end, std::back_inserter(ret),
+                 [&idx, op = std::forward<UnaryOp>(op)](auto& item) mutable {
+                   return op(idx++, item);
+                 });
+  return ret;
+}
+
+template <typename Container, typename UnaryOp>
+auto PmrMap(Container&& container, UnaryOp&& op) {
+  return PmrMap(std::begin(container), std::end(container),
+                std::forward<UnaryOp>(op));
 }
 
 template <typename Iterator, typename UnaryOp,
--- a/tachyon/base/json/rapidjson_util.h
+++ b/tachyon/base/json/rapidjson_util.h
@@ -1,6 +1,7 @@
 #ifndef TACHYON_BASE_JSON_RAPIDJSON_UTIL_H_
 #define TACHYON_BASE_JSON_RAPIDJSON_UTIL_H_
 
+#include <memory_resource>
 #include <optional>
 #include <string>
 #include <string_view>
@@ -237,12 +238,11 @@ class RapidJsonValueConverter<T, std::enable_if_t<std::is_enum<T>::value>> {
   }
 };
 
-template <typename T>
-class RapidJsonValueConverter<std::vector<T>> {
+template <typename V, typename T = typename V::value_type>
+class RapidJsonValueConverterVectorImpl {
  public:
   template <typename Allocator>
-  static rapidjson::Value From(const std::vector<T>& value,
-                               Allocator& allocator) {
+  static rapidjson::Value From(const V& value, Allocator& allocator) {
     rapidjson::Value array(rapidjson::kArrayType);
     for (size_t i = 0; i < value.size(); ++i) {
       array.PushBack(RapidJsonValueConverter<T>::From(value[i], allocator),
@@ -252,12 +252,12 @@ class RapidJsonValueConverter<std::vector<T>> {
   }
 
   static bool To(const rapidjson::Value& json_value, std::string_view key,
-                 std::vector<T>* value, std::string* error) {
+                 V* value, std::string* error) {
     if (!json_value.IsArray()) {
       *error = RapidJsonMismatchedTypeError(key, "array", json_value);
       return false;
     }
-    std::vector<T> value_tmp;
+    V value_tmp;
     for (auto it = json_value.Begin(); it != json_value.End(); ++it) {
       T v;
       if (!RapidJsonValueConverter<T>::To(*it, key, &v, error)) return false;
@@ -267,6 +267,13 @@ class RapidJsonValueConverter<std::vector<T>> {
     return true;
   }
 };
+
+template <typename T>
+class RapidJsonValueConverter<std::vector<T>>
+    : public RapidJsonValueConverterVectorImpl<std::vector<T>> {};
+template <typename T>
+class RapidJsonValueConverter<std::pmr::vector<T>>
+    : public RapidJsonValueConverterVectorImpl<std::pmr::vector<T>> {};
 
 template <typename T>
 class RapidJsonValueConverter<std::optional<T>> {
--- a/tachyon/c/zk/plonk/halo2/buffer_reader.h
+++ b/tachyon/c/zk/plonk/halo2/buffer_reader.h
@@ -2,6 +2,7 @@
 #define TACHYON_C_ZK_PLONK_HALO2_BUFFER_READER_H_
 
 #include <memory>
+#include <memory_resource>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -106,6 +107,16 @@ class BufferReader<std::vector<T>> {
  public:
   static std::vector<T> Read(const tachyon::base::ReadOnlyBuffer& buffer) {
     return tachyon::base::CreateVector(ReadU32AsSizeT(buffer), [&buffer]() {
+      return BufferReader<T>::Read(buffer);
+    });
+  }
+};
+
+template <typename T>
+class BufferReader<std::pmr::vector<T>> {
+ public:
+  static std::pmr::vector<T> Read(const tachyon::base::ReadOnlyBuffer& buffer) {
+    return tachyon::base::CreatePmrVector(ReadU32AsSizeT(buffer), [&buffer]() {
       return BufferReader<T>::Read(buffer);
     });
   }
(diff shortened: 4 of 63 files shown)


tachyon/base/parallelize.h
```diff
@@ -105,6 +105,53 @@ auto ParallelizeMap(Container& container, Callable callback,
...
+// See parallelize_unittest.cc for more details.
+template <typename Container, typename Callable,
+          typename FunctorTraits = internal::MakeFunctorTraits<Callable>,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          typename SpanTy = internal::GetType<0, ArgList>,
+          typename T = typename SpanTy::value_type,
+          size_t ArgNum = internal::GetSize<ArgList>>
+std::pmr::vector<ReturnType> PmrParallelizeMapByChunkSize(Container& container,
```
chokobole, line 120: Please remove `PmrParallelizeMapByChunkSize()` and `PmrParallelizeMap()`. The size of the vector will be the number of threads, which typically is less than 64, so there is no merit in keeping these functions.

TomTaehoonKim (approved):
LGTM

tachyon/math/base/semigroups.h
```diff
@@ -215,6 +215,22 @@ class MultiplicativeSemigroup {
     return ret;
   }
 
+  constexpr static std::pmr::vector<MulResult> GetSuccessivePowersPmr(
```
chokobole, line 218: What about keeping only one `GetSuccessivePowers()` that uses `std::pmr::vector`? If we have two versions, users might get confused. Additionally, this typically generates long series of powers, so having a single, efficient implementation would be beneficial.

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
```diff
@@ -235,6 +235,13 @@ class UnivariateEvaluationDomain : public EvaluationDomain<F, MaxDegree> {
     return F::GetSuccessivePowers(size, root);
   }
 
+  // Computes the first |size| roots of unity for the entire domain.
+  // e.g. for the domain [1, g, g², ..., gⁿ⁻¹}] and |size| = n / 2, it
+  // computes [1, g, g², ..., g^{(n / 2) - 1}]
+  constexpr std::pmr::vector<F> GetRootsOfUnityPmr(size_t size, const F& root) const {
+    return F::GetSuccessivePowersPmr(size, root);
+  }
```
chokobole: What about keeping only one `GetRootsOfUnity()` that uses `std::pmr::vector`? If we have two versions, users might get confused. Additionally, this typically generates long series of powers, so having a single, efficient implementation would be beneficial.

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
```diff
@@ -238,7 +238,8 @@ class UnivariateEvaluationDomain : public EvaluationDomain<F, MaxDegree> {
   // Computes the first |size| roots of unity for the entire domain.
   // e.g. for the domain [1, g, g², ..., gⁿ⁻¹}] and |size| = n / 2, it
   // computes [1, g, g², ..., g^{(n / 2) - 1}]
-  constexpr std::pmr::vector<F> GetRootsOfUnityPmr(size_t size, const F& root) const {
+  constexpr std::pmr::vector<F> GetRootsOfUnityPmr(size_t size,
+                                                   const F& root) const {
```
chokobole, line 242: This should be squashed into the previous commit.

tachyon/zk/r1cs/constraint_system/qap_witness_map_result.h
```diff
@@ -1,13 +1,14 @@
 #ifndef TACHYON_ZK_R1CS_CONSTRAINT_SYSTEM_QAP_WITNESS_MAP_RESULT_H_
 #define TACHYON_ZK_R1CS_CONSTRAINT_SYSTEM_QAP_WITNESS_MAP_RESULT_H_
 
+#include <memory_resource>
```
chokobole, line 4: Please include `<memory_resource>` to all the headers and the source files whose headers don't include this.

tachyon/math/base/semigroups.h
batzor: @chokobole  The scope is `math` so I thought there can be other use cases for this as well.

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
batzor: It is used in [`radix2_evaluation_domain`](https://github.com/kroma-network/tachyon/blob/e8558ecddb6b36dfc0d63e78cf5e8c940844eb06/tachyon/math/polynomials/univariate/radix2_evaluation_domain.h#L300). So the change can propagate to packed prime fields as well.

tachyon/base/parallelize.h
batzor: `PmrParallelizeMap()` is used in [`DoEvaluate`](https://github.com/kroma-network/tachyon/blob/e8558ecddb6b36dfc0d63e78cf5e8c940844eb06/tachyon/math/polynomials/univariate/univariate_dense_coefficients.h#L267)

tachyon/math/base/semigroups.h
batzor: I was thinking of cases when we use small size. But I guess we don't care about performance much in those cases. I will remove it in separate commit since it will be big change.

batzor:
@TomTaehoonKim Re-requesting review since I included some additional changes.

tachyon/c/zk/plonk/halo2/buffer_reader.h
```diff
@@ -201,7 +201,7 @@ class BufferReader<tachyon::math::UnivariateDensePolynomial<F, MaxDegree>> {
  public:
   static tachyon::math::UnivariateDensePolynomial<F, MaxDegree> Read(
       const tachyon::base::ReadOnlyBuffer& buffer) {
-    std::vector<F> coeffs;
+    std::pmr::vector<F> coeffs;
```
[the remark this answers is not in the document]
batzor: It already has it from previous commit.

tachyon/base/containers/container_util.h
```diff
@@ -97,6 +98,42 @@ std::vector<ReturnType> CreateVectorParallel(size_t size,
...
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 0>* = nullptr>
+std::pmr::vector<ReturnType> CreatePmrVector(size_t size,
+                                             Generator&& generator) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(size);
+  std::generate_n(std::back_inserter(ret), size,
+                  std::forward<Generator>(generator));
```
GideokKim, line 113: Is this code inefficient?

```suggestion
  std::pmr::vector<ReturnType> ret(size);
  std::generate_n(ret.begin(), size, std::forward<Generator>(generator));
```

tachyon/base/buffer/copyable.h
```diff
@@ -162,6 +163,36 @@ class Copyable<std::vector<T>> {
...
+  }
+
+  static size_t EstimateSize(const std::pmr::vector<T>& values) {
+    return std::accumulate(values.begin(), values.end(), sizeof(size_t),
+                           [](size_t total, const T& value) {
+                             return total + base::EstimateSize(value);
+                           });
+  }
+};
+
```
dongchangYoo, line 195: The code for std::vector and std::pmr::vector is identical for the Copyable aspect. Can we increase code reuse?

dongchangYoo (changes requested):

dongchangYoo:
I have a question. Are there any issues that might arise from replacing all instances of std::vector with std::pmr::vector?

tachyon/base/containers/container_util.h
batzor: This will actually construct `size` number of elements which is equivalent of `resize()`. But here we just `reserve()`.

tachyon/base/json/rapidjson_util.h
```diff
@@ -268,6 +268,37 @@ class RapidJsonValueConverter<std::vector<T>> {
   }
 };
 
+template <typename T>
+class RapidJsonValueConverter<std::pmr::vector<T>> {
```
GideokKim, line 272: Please include `<memory_resource>`
Please also check the files below:
`tachyon/math/polynomials/univariate/univariate_dense_polynomial_unittest.cc`
`tachyon/math/polynomials/univariate/univariate_evaluation_domain_unittest.cc`
`tachyon/math/polynomials/univariate/univariate_evaluations_unittest.cc`
`tachyon/zk/plonk/constraint_system/selector_compressor_unittest.cc`
`tachyon/zk/plonk/expressions/compress_expression_unittest.cc`
`tachyon/zk/plonk/permutation/unpermuted_table_unittest.cc`
`vendors/circom/prover_main.cc`

batzor: Generally, I didn't include in the headers which doesn't include `<vector>` in the first place.

batzor:
@dongchangYoo `std::pmr::vector` can be slow in small vectors and we might replace `std::pmr::vector` for something more performant such as `std::vector<T, VeryOptimalAllocator<T>>`. You can think of this as intermediate step of optimizing large vectors. I think implementing `VeryOptimalAllocator<T>` should be done later as a separate PR later on because it will require a lot of testing and benchmarking. One example is here: https://github.com/bitcoin/bitcoin/pull/25325.

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
```diff
@@ -138,10 +139,32 @@ class UnivariateEvaluationDomain : public EvaluationDomain<F, MaxDegree> {
...
+          CHECK(
+              (*icicle_)->FFT(FFTAlgorithmToIcicleNTTAlgorithm(GetAlgorithm()),
+                              offset_big_int_, evals));
+          return evals;
+        }
+      }
+#endif
+      DoFFT(evals);
+    }
+    return evals;
```
GideokKim, line 159: Similar to IFFT, wouldn't it be good to return first in the case of zero?

```suggestion
    if (poly.IsZero()) return evals;

    evals.evaluations_ = poly.coefficients_.coefficients_;
#if TACHYON_CUDA
    // TODO(chokobole): Remove this condition.
    if constexpr (std::is_same_v<F, bn254::Fr>) {
      if (icicle_) {
        evals.evaluations_.resize(size_);
        CHECK((*icicle_)->FFT(FFTAlgorithmToIcicleNTTAlgorithm(GetAlgorithm()),
                              offset_big_int_, evals));
        return evals;
      }
    }
#endif
    DoFFT(evals);

    return evals;
```

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
```diff
@@ -138,10 +139,32 @@ class UnivariateEvaluationDomain : public EvaluationDomain<F, MaxDegree> {
...
+  }
+
+  // Compute a FFT.
+  [[nodiscard]] constexpr Evals FFT(DensePoly&& poly) const {
     if (poly.IsZero()) return {};
 
     Evals evals;
```
GideokKim, line 166: It looks like optimization is possible here too, is that correct?

```suggestion
    Evals evals;
    if (poly.IsZero()) return evals;
```

tachyon/math/polynomials/univariate/univariate_evaluation_domain.h
batzor: Performance-wise, it won't have any difference since in the case of 0, it will not enter the `if` branch. But I will change it for consistency.

GideokKim (approved):
LGTM

tachyon/base/buffer/copyable.h
batzor: reflected.

tachyon/base/json/rapidjson_util.h
batzor: I only added to rapidjson. Others rely on transitive include in the first place so I didn't add include.

tachyon/math/polynomials/univariate/lagrange_interpolation.h
```diff
@@ -63,7 +64,7 @@ bool LagrangeInterpolate(
         // clang-format on
 
         std::vector<F> coeffs(points.size());
```
ashjeong, line 66: Should this also be a `pmr::vector`?

tachyon/c/zk/plonk/halo2/kzg_family_prover_impl.h
```diff
@@ -30,18 +31,19 @@ class KZGFamilyProverImpl : public ProverImplBase<PCS, LS> {
...
   }
 
   CProjectivePoint* CommitLagrangeRaw(
-      const std::vector<ScalarField>& scalars) const {
+      const std::pmr::vector<ScalarField>& scalars) const {
     return DoMSM(this->pcs_.GetG1PowersOfTauLagrange(), scalars);
   }
 
  private:
   CProjectivePoint* DoMSM(const std::vector<AffinePoint>& bases,
```
ashjeong, line 45: Should this also be a `pmr::vector`?

tachyon/math/base/semigroups.h
ashjeong: Can this be moved to [the commit two after](https://github.com/kroma-network/tachyon/pull/486/commits/cda7427f921e31fb8c4144a784579f2a1a00c830) in order, since the commit is about this replacement anyway?

tachyon/zk/r1cs/constraint_system/quadratic_arithmetic_program.h
```diff
@@ -126,9 +126,11 @@ class QuadraticArithmeticProgram {
 
     CHECK_GE(domain->size(), matrices.num_constraints);
 
-    std::pmr::vector<F> a(domain->size());
-    std::pmr::vector<F> b(domain->size());
-    std::pmr::vector<F> c(domain->size());
+    std::vector<Evals> abc;
+    abc.reserve(3);
```
ashjeong, line 130:
```suggestion
    std::vector<Evals> abc(3);

File: vendors/circom/circomlib/circuit/quadratic_arithmetic_program.h [hunk 15]
```diff
@@ -28,9 +28,11 @@ class QuadraticArithmeticProgram {
     using Evals = typename Domain::Evals;
     using DensePoly = typename Domain::DensePoly;
 
-    std::pmr::vector<F> a(domain->size());
-    std::pmr::vector<F> b(domain->size());
-    std::pmr::vector<F> c(domain->size());
+    std::vector<Evals> abc;
+    abc.reserve(3);
```
ashjeong, line 32:
```suggestion
    std::vector<Evals> abc(3);

File: tachyon/base/containers/container_util.h [hunk 16]
```diff
@@ -97,6 +98,42 @@ std::vector<ReturnType> CreateVectorParallel(size_t size,
...
+          typename FunctorTraits = internal::MakeFunctorTraits<Generator>,
+          typename ReturnType = typename FunctorTraits::ReturnType,
+          typename RunType = typename FunctorTraits::RunType,
+          typename ArgList = internal::ExtractArgs<RunType>,
+          size_t ArgNum = internal::GetSize<ArgList>,
+          std::enable_if_t<ArgNum == 0>* = nullptr>
+std::pmr::vector<ReturnType> CreatePmrVector(size_t size,
+                                             Generator&& generator) {
+  std::pmr::vector<ReturnType> ret;
+  ret.reserve(size);
```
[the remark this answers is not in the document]
batzor: This will actually construct size number of elements which is equivalent of resize(). But here we just reserve().

tachyon/math/base/semigroups.h
batzor: It won't compile in the next commit in that case.

tachyon/zk/r1cs/constraint_system/quadratic_arithmetic_program.h
batzor: This will construct evals as if I did `abc.resize(3)`. Behavior is different from `reserve()`.

vendors/circom/circomlib/circuit/quadratic_arithmetic_program.h
batzor: This will construct evals as if I did `abc.resize(3)`. Behavior is different from `reserve()`. And following `emplace_back`s would make the size 6.

dongchangYoo (approved):
LGTM

tachyon/math/base/semigroups.h
```diff
@@ -197,25 +198,7 @@ class MultiplicativeSemigroup {
...
-          if (!c.IsOne()) pow *= c;
-          for (G& v : chunk) {
-            v = pow;
-            pow *= generator;
-          }
-        });
-    return ret;
-  }
-
-  constexpr static std::pmr::vector<MulResult> GetSuccessivePowersPmr(
```
TomTaehoonKim: Can you get rid of this intermediate change?

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -223,6 +232,43 @@ class Radix2EvaluationDomain : public UnivariateEvaluationDomain<F, MaxDegree>,
     OutInHelper(evals, start_gap);
   }
 
+  CONSTEXPR_IF_NOT_OPENMP void DegreeAwareFFTInPlace(
+      std::vector<Evals>& evalss) const {
```
TomTaehoonKim, line 236:
```suggestion
      std::vector<Evals>& evals) const {
```

tachyon/math/polynomials/univariate/batch_fft_benchmark.cc
```diff
@@ -8,26 +8,30 @@ namespace tachyon::math::polynomials::univariate {
 
 constexpr size_t kMaxDegree = (size_t{1} << 32) - 1;
 template <typename F>
-void BM_BatchFFTSerial(benchmark::State& state) {
+void BM_SerialFFT(benchmark::State& state) {
```
TomTaehoonKim, line 11: Can avoid intermediate change?

tachyon/math/base/semigroups.h
batzor: I think that would be not so simple. Changing the `GetSuccessivePowers` to `std::pmr::vector` by default would cause the `UnivariateEvaluations` to use pmr vector.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
batzor: I was also not so sure about what to use here. `evals` is used for single instance of `Evals` across the repository. Do you think it's fine to use `evals` for vector and `eval` for single instance here?

tachyon/math/polynomials/univariate/batch_fft_benchmark.cc
batzor: Sorry this commit should be squashed with the original. So it will be removed.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
TomTaehoonKim: Hmm but technically evals is correct for the single instance so let's just use evalss for vector of evals.

TomTaehoonKim (approved):
LGTM

tachyon/math/base/semigroups.h
```diff
@@ -215,6 +215,22 @@ class MultiplicativeSemigroup {
...
+      size_t size, const G& generator, const G& c = G::One()) {
+    std::pmr::vector<MulResult> ret(size);
+    base::Parallelize(
+        ret, [&generator, &c](absl::Span<G> chunk, size_t chunk_offset,
+                              size_t chunk_size) {
+          MulResult pow = generator.Pow(chunk_offset * chunk_size);
+          if (!c.IsOne()) pow *= c;
+          for (G& v : chunk) {
+            v = pow;
+            pow *= generator;
```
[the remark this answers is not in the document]
chokobole: Please add `CHECK(!chunk.empty())` since if the `chunk.size() == 0` it will run through very long loop.

Please apply this trick to `GetSuccessiveScalarMuls()`, `EvaluatePartialLagrangeCoefficients()`, `DivideByVanishingPolyInPlace()` and `ComputeHQuery()`, too!

tachyon/math/base/semigroups.h
```diff
@@ -206,10 +206,11 @@ class MultiplicativeSemigroup {
                               size_t chunk_size) {
           MulResult pow = generator.Pow(chunk_offset * chunk_size);
           if (!c.IsOne()) pow *= c;
-          for (G& v : chunk) {
-            v = pow;
+          for (size_t i = 0; i < chunk.size() - 1; ++i) {
+            chunk[i] = pow;
             pow *= generator;
           }
+          chunk.back() = pow;
```
chokobole, line 213:
```suggestion
          chunk.back() = std::move(pow);
```

tachyon/math/base/semigroups.h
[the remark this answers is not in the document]
batzor: I don't think it's possible to have empty chunk in `base::Parallelize`

chokobole (approved):
LGTM

ashjeong (approved):
LGTM

PR #491, opened 2024-07-25
perf/optimize-prime-field-inversion-bernstein-yang-method

AlekseiVambol:
Replace for prime fields the multiplicative inversion based on the Binary Extended Euclidean method with the Bernstein-Yang multiplicative inversion. The latter one turned out to be more than 3 times faster on average.
image

### DIFF
--- a/tachyon/math/base/byinverter.h
+++ b/tachyon/math/base/byinverter.h
@@ -1,0 +1,449 @@
+#ifndef TACHYON_MATH_BASE_BYINVERTER_H_
+#define TACHYON_MATH_BASE_BYINVERTER_H_
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#include <algorithm>
+
+#include "absl/numeric/int128.h"
+
+#include "tachyon/base/bits.h"
+#include "tachyon/math/base/big_int.h"
+
+namespace tachyon::math {
+
+// Unsigned 128-bit integer type
+typedef absl::uint128 uint128_t;
+
+// Big signed (B * L)-bit integer type, whose variables
+// store numbers in the two's complement code as arrays
+// of B-bit chunks, where B < 64 and L > 0. The ordering
+// of the chunks in these arrays is little-endian. The
+// arithmetic operations for this type are wrapping ones
+template <size_t B, size_t L>
+class CInt {
+ public:
+  // Mask, in which the B lowest bits are 1 and only they
+  constexpr static uint64_t MASK = ~uint64_t{0} >> (64 - B);
+
+  // Returns the number equal to the argument. The binary
+  // representation of the absolute value of the argument
+  // must have no more than B * L - 1 bits to ensure that
+  // the result and the argument represent the same number
+  constexpr static CInt Load(const int64_t input) {
+    CInt result;
+    uint64_t raw = input;
+    Convert<64, B>(&raw, 1, result.data, L);
+    if ((B * L > 64) && (input < 0)) {
+      // For the two's complement code and n > m the n-bit representation
+      // of a negative number can be expressed as its m-bit representation
+      // preceded by m - n bits equal to 1
+      result.data[64 / B] |= (~uint64_t{0} << (64 % B)) & MASK;
+      for (size_t i = 64 / B + 1; i < L; i++) {
+        result.data[i] = MASK;
+      }
+    }
+    return result;
+  }
+
+  // Returns the number, the absolute value of which is specified by the
+  // first argument. A non-zero return number is negative iff "sign" is
+  // "true". The binary representation of the absolute value of the input
+  // number must have no more than B * L - 1 bits to ensure that the result
+  // and arguments specify the same number
+  template <size_t N>
+  constexpr static CInt Load(const BigInt<N> &input, bool sign) {
+    CInt result;
+    Convert<64, B>(input.limbs, N, result.data, L);
+    return sign ? -result : result;
+  }
+
+  // Returns the current number. The binary representation
+  // of its absolute value must have no more than 63 bits
+  // to ensure that the result represents the current number
+  constexpr int64_t Save() const {
+    uint64_t result = 0;
+    Convert<B, 64>(data, L, &result, 1);
+    if ((B * L < 64) && IsNegative()) {
+      // For the two's complement code and n > m the n-bit representation
+      // of a negative number can be expressed as its m-bit representation
+      // preceded by m - n bits equal to 1
+      result |= ~uint64_t{0} << (B * L);
+    }
+    return result;
+  }
+
+  // Returns the sign of the current number and stores its absolute
+  // value into the variable, which is specified by the argument.
+  // The binary representation of the absolute value of the current
+  // number must contain no more than 64 * N bits to ensure that the
+  // result represents the current number
+  template <size_t N>
+  constexpr bool Save(BigInt<N> &output) const {
+    bool sign = IsNegative();
+    Convert<B, 64>((sign ? -*this : *this).data, L, output.limbs, N);
+    return sign;
+  }
+
+  // Returns the lowest B bits of the current number
+  constexpr uint64_t Lowest() const { return data[0]; }
+
+  // Returns "true" iff the current number is negative
+  constexpr bool IsNegative() const { return data[L - 1] > (MASK >> 1); }
+
+  // Returns the result of applying B-bit right
+  // arithmetical shift to the current number
+  constexpr CInt Shift() const {
+    CInt result;
+    if (IsNegative()) {
+      result.data[L - 1] = MASK;
+    }
+    for (size_t i = 1; i < L; i++) {
+      result.data[i - 1] = data[i];
+    }
+    return result;
+  }
+
+  constexpr bool operator==(const CInt &other) const {
+    for (size_t i = 0; i < L; i++) {
+      if (data[i] != other.data[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  constexpr bool operator!=(const CInt &other) const {
+    return !(*this == other);
+  }
+
+  constexpr CInt operator+(const CInt &other) const {
+    CInt result;
+    uint64_t carry = 0;
+    for (size_t i = 0; i < L; i++) {
+      uint64_t sum = data[i] + other.data[i] + carry;
+      result.data[i] = sum & MASK;
+      carry = sum >> B;
+    }
+    return result;
+  }
+
+  constexpr CInt operator-(const CInt &other) const {
+    // For the two's complement code the additive negation is the result of
+    // adding 1 to the bitwise inverted argument's representation. Thus, for
+    // any encoded integers x and y we have x - y = x + !y + 1, where "!" is
+    // the bitwise inversion and addition is done according to the rules of
+    // the code. The algorithm below uses this formula and is the modified
+    // addition algorithm, where the carry flag is initialized with 1 and
+    // the chunks of the second argument are bitwise inverted
+    CInt result;
+    uint64_t carry = 1;
+    for (size_t i = 0; i < L; i++) {
+      uint64_t sum = data[i] + (other.data[i] ^ MASK) + carry;
+      result.data[i] = sum & MASK;
+      carry = sum >> B;
+    }
+    return result;
+  }
+
+  constexpr CInt operator-() const {
+    // For the two's complement code the additive negation is the result
+    // of adding 1 to the bitwise inverted argument's representation
+    CInt result;
+    uint64_t carry = 1;
+    for (size_t i = 0; i < L; i++) {
+      uint64_t sum = (data[i] ^ MASK) + carry;
+      result.data[i] = sum & MASK;
+      carry = sum >> B;
+    }
+    return result;
+  }
+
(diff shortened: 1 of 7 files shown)


batzor:
Hi @AlekseiVambol! Please squash your commits to fewer commits and your commit message should follow the guideline to summarize what the commit does. You can refer to other PRs as well.

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  /// Returns the number equal to the one specified by the arguments. Its
+  /// absolute value is (64 * "isize")-bit integer stored as the array of
+  /// 64-bit chunks, whose address is specified by "input". The ordering
+  /// of the chunks in the array is little-endian. A non-zero input number
+  /// is negative iff "sign" is "true". The binary representation of the
+  /// absolute value of the input number must have no more than B * L - 1
+  /// bits to ensure that the result and arguments specify the same number
+  constexpr static CInt load(const u64 *input, const usize isize, bool sign) {
```
chokobole, line 59: Please use `BigInt<N>` instead of `input` and `isize`, also include the boundary check at compile time.

```
  template <typename N>
  constexpr static CInt load(const BigInt<N>& input, bool sign) {
    static_assert(N * 64 <= B * L);
  }
```

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+    return sign ? -result : result;
+  }
+
+  /// Returns the current number. The binary representation
+  /// of its absolute value must have no more than 63 bits
+  /// to ensure that the result represents the current number
+  constexpr i64 save() const {
+    u64 result = 0;
+    CInt::convert<B, 64>(this->data, L, &result, 1);
+    if ((B * L < 64) && this->is_negative()) {
```
chokobole, line 71:
```suggestion
    CInt::convert<B, 64>(data, L, &result, 1);
    if ((B * L < 64) && is_negative()) {
```

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+/// Returns the number of trailing zeros in "n"
+/// represented in the little-endian binary code
+constexpr usize trailing0(u64 n) { return n != 0 ? __builtin_ctzll(n) : 64; }
+
+/// Big signed (B * L)-bit integer type, whose variables
+/// store numbers in the two's complement code as arrays
+/// of B-bit chunks, where B < 64 and L > 0. The ordering
+/// of the chunks in these arrays is little-endian. The
+/// arithmetic operations for this type are wrapping ones
+template <usize B, usize L> struct CInt {
```
chokobole: In most cases, `struct` uses for a container that has public member variables, so in this case, you should change it to `class`.

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+        u128 sum =
+            (u128)this->data[i] * other.data[k - i] + result.data[k] + carry;
+        result.data[k] = (u64)(sum & CInt::MASK);
+        carry = (u64)(sum >> B);
+      }
+    }
+    return result;
+  }
+
+  CInt operator*(i64 other) const {
```
chokobole, line 182:
```suggestion
  constexpr CInt operator*(i64 other) const {
```

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+
+  constexpr CInt operator*(const CInt &other) const {
+    CInt result;
+    for (usize i = 0; i < L; i++) {
+      u64 carry = 0;
+      for (usize k = i; k < L; k++) {
+        u128 sum =
+            (u128)this->data[i] * other.data[k - i] + result.data[k] + carry;
+        result.data[k] = (u64)(sum & CInt::MASK);
+        carry = (u64)(sum >> B);
```
chokobole, line 176:
```suggestion
        result.data[k] = static_cast<uint64_t>(sum & CInt::MASK);
        carry = static_cast<uint64_t>(sum >> B);
```

chokobole:
Please add unittest for `byinverter.h`

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+  /// absolute value of the current number must contain at most 64 * "osize"
+  /// bits to ensure that the result represents the current number
+  constexpr bool save(u64 *output, const usize osize) const {
+    bool sign = this->is_negative();
+    CInt::convert<B, 64>((sign ? -*this : *this).data, L, output, osize);
+    return sign;
+  }
+
+  /// Returns the lowest B bits of the current number
+  constexpr u64 lowest() const { return this->data[0]; }
```
chokobole, line 93: Please avoid using `this->` if you access members declared in the same class!

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+
+  /// Stores at the address "output" the array of O-bit chunks, which represents
+  /// the big unsigned integer equal modulo 2 ^ (O * "osize") to the input big
+  /// unsigned integer stored at the address "input" as an array of "isize"
+  /// I-bit chunks. The ordering of the chunks in these arrays is little-endian
+  template <usize I, usize O>
+  constexpr static void convert(const u64 *input, const usize isize,
+                                u64 *output, const usize osize) {
+    usize bits = 0;
+    usize total = least(isize * I, osize * O);
```
chokobole, line 225: Why do you allow the case that output size is less than input size?

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+      usize o = bits % O;
+      output[bits / O] |= (input[bits / I] >> i) << o;
+      bits += least(I - i, O - o);
+    }
+    u64 mask = ~0UL >> (64 - O);
+    usize filled = (total + O - 1) / O;
+    while (filled > 0) {
+      filled -= 1;
+      output[filled] &= mask;
+    }
```
chokobole, line 240:
```suggestion
    for (size_t i = filled - 1; i != SIZE_MAX; --i) {
      output[i] &= mask;
    }
```

tachyon/math/base/byinverter.h
chokobole: Just a question, why do you name it to `CInt`? what does this stand for?

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+    if (negate) {
+      value = -value;
+    }
+
+    if (value.is_negative()) {
+      value = value + this->modulus;
+    }
+  }
+
+  /// Returns the multiplicative inverse of the argument modulo 2^62. The
```
chokobole, line 411: Please use unicode for exponent 2⁶²

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+    d = cd.shift();
+    e = ce.shift();
+  }
+
+  /// Stores in the variable "value" either "value (mod M)" or
+  /// "-value (mod M)", where M is the modulus the inverter was
+  /// created for, depending on "negate", which determines the
+  /// presence of "-" in the used formula. The initial integer
+  /// contained in "value" lies in (-2 * M, M)
+  void norm(LCInt &value, const bool negate) const {
```
chokobole, line 397: Please make the function return the result of the function, e.g `Inverse(x) -> x⁻¹`

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
+#ifndef TACHYON_MATH_BASE_BYINVERTER_H_
+#define TACHYON_MATH_BASE_BYINVERTER_H_
+
+namespace tachyon::math {
+/// Signed 64-bit integer type
+typedef __int64_t i64;
+/// Unsigned 64-bit integer type
+typedef __uint64_t u64;
+/// Unsigned 128-bit integer type
+typedef __uint128_t u128;
```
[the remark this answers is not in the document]
AlekseiVambol: absl::uint128 may have bad impact on performance. It is class, not the primitive intrinsic type. Both clang and gcc with c++17 option support  __uint128_t. No need in absl. In general, I was aimed at creating the code which is independent on any libraries, especially non-standard one, to have better control over performance.

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+/// Unsigned integer type for storing object sizes
+typedef size_t usize;
+
+/// Returns the non-largest argument
+constexpr usize least(usize a, usize b) { return a < b ? a : b; }
+
+/// Returns the number of trailing zeros in "n"
+/// represented in the little-endian binary code
+constexpr usize trailing0(u64 n) { return n != 0 ? __builtin_ctzll(n) : 64; }
```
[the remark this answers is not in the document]
AlekseiVambol: Ok. Influence of the LIKELY macro in its code may be beneficial.

tachyon/math/base/byinverter.h
```diff
@@ -0,0 +1,469 @@
...
+        adjuster(LCInt::load(adjuster, asize, false)),
+        inverse62(BYInverter::invert62(modulus[0])) {}
+
+  /// Returns "true" and stores in the variable, which is specified by the
+  /// second argument pair, the adjusted modular multiplicative inverse of
+  /// the input number, which is specified by the first argument pair, if
+  /// the input number is invertible for the modulus of the invertor (i.e.
+  /// coprime with it). Otherwise, "false" is returned
+  bool invert(const u64 *input, const usize isize, u64 *output,
+              const usize osize) const {
```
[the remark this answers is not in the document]
AlekseiVambol: I will be less flexible, but ok.

tachyon/math/base/byinverter.h
AlekseiVambol: Custom (or Chunked) Integer.

tachyon/math/base/byinverter.h
AlekseiVambol: Sometimes only part of the input is needed, i.e. here
image
Sometimes less input bits are available than can completely occupy the output array, and the last one should be completely defined by setting its last bits to zero. An example (for a positive input) 
image

tachyon/math/base/byinverter.h
AlekseiVambol: Your BigInt is unsigned, isn't it?

tachyon/math/base/byinverter.h
AlekseiVambol: However, for people, who do not know the two's complement code, the new version will be obscure. Should I leave it as is?

AlekseiVambol: Let it be 
image

tachyon/math/base/byinverter.h
AlekseiVambol: `  template <size_t N>
  constexpr static CInt Load(const BigInt<N> &input, bool sign) {
    CInt result;
    Convert<64, B>(input.limbs, N, result.data, L);
    return sign ? -result : result;
  }`

For this data type I would like to allow loading from the larger BigInts: some using parties may know that certain BigInt variable may contain some small value due to features of an algorithm.

tachyon/math/base/byinverter.h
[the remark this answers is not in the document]
AlekseiVambol: I introduced absl::uint128 in the code, because this has not decreased the performance.

AlekseiVambol:
> Please add unittest for `byinverter.h`

Added.

AlekseiVambol:
> Hi @AlekseiVambol! Please squash your commits to fewer commits and your commit message should follow the [guideline](https://github.com/kroma-network/tachyon/blob/d36769b5f0ce5551c7e0c492713ff509e58de9ec/.gitmessage) and summarize what the commit does. You can refer to other PRs as well.

Done.

AlekseiVambol:
Seems to be ready for merging. Now it is even faster: average performance gain has increased from 3x to 3.5x.
image

tachyon/math/finite_fields/prime_field_fallback.h
```diff
@@ -295,20 +299,18 @@ class PrimeField<_Config, std::enable_if_t<!_Config::kUseAsm &&
   // MultiplicativeGroup methods
   constexpr std::optional<PrimeField> Inverse() const {
     PrimeField ret{};
-    if (LIKELY(value_.template MontgomeryInverse<Config::kModulusHasSpareBit>(
-            Config::kModulus, Config::kMontgomeryR2, ret.value_))) {
+    if (inverter.Invert(value_, ret.value_)) {
       return ret;
     }
-    LOG_IF_NOT_GPU(ERROR) << "Inverse of zero attempted";
+    LOG_IF_NOT_GPU(ERROR) << "Inversion of zero attempted";
```
chokobole, line 305: Please restore this change! there's another place to use this error message, it's better to add this change in a separate PR if it is needed.

chokobole:
Please follow the https://github.com/kroma-network/.github/blob/main/CONTRIBUTING.md#commit-messages!

AlekseiVambol:
Ready.

chokobole (approved):
LGTM

PR #498, opened 2024-07-30
ci: fix warnings and improve runtime

batzor:
# Description

This PR fixes unused variable warnings and improves CI runtime from ~50 minutes to ~30 minutes. Also, `exclusive` tag is introduced to cpu heavy tests to prevent timeouts from happening. 

Closes #197 since it's no longer reproducible.

PR #503, opened 2024-08-02
test: optimize openmp test

chokobole:
# Description

#498 introduced an `exclusive` tag that caused some tests to run sequentially, which slowed down the test times. This PR improves the situation by limiting the number of threads to 4. This change increases the likelihood of passing the unit tests and reduces the overall test time.

See https://bazel.build/reference/be/common-definitions#common.tags.

### DIFF
--- a/tachyon/zk/plonk/examples/BUILD.bazel
+++ b/tachyon/zk/plonk/examples/BUILD.bazel
@@ -1,4 +1,9 @@
-load("//bazel:tachyon_cc.bzl", "tachyon_cc_library", "tachyon_cc_unittest")
+load(
+    "//bazel:tachyon_cc.bzl",
+    "tachyon_cc_library",
+    "tachyon_cc_unittest",
+    "tachyon_openmp_num_threads_env",
+)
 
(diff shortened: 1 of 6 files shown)


TomTaehoonKim (approved):
LGTM

tachyon/zk/plonk/examples/BUILD.bazel
```diff
@@ -171,7 +176,6 @@ tachyon_cc_unittest(
 tachyon_cc_unittest(
     name = "simple_circuit_test",
     srcs = ["simple_circuit_test.cc"],
-    tags = ["exclusive"],
```
GideokKim, line 174: Is the original intention of this test to only delete tags?

dongchangYoo (approved):
LGTM

ashjeong (approved):
LGTM

batzor (approved):
I was looking for another solution without limiting the threads but it seems be hard.

LGTM

GideokKim (approved):
LGTM

PR #515, opened 2024-08-09
feat: add `FFTBatch` benchmark

batzor:
# Description

This PR adds `FFTBatch` benchmark of Tachyon and Plonky3.

PR #517, opened 2024-08-12
perf: optimize fft batch

chokobole:
# Description

PR #515 introduces an FFT batch benchmark. Initially, the results were not as expected, but after some optimizations, they improved. However, there's still potential for further enhancements. Given that our implementation precomputes twiddles, we would expect even better performance. We need to investigate this further to understand the underlying reasons and explore additional optimizations.


batzor (approved):
LGTM

GideokKim (approved):
LGTM

TomTaehoonKim (approved):
LGTM

dongchangYoo (approved):
LGTM

ashjeong (approved):
LGTM

PR #525, opened 2024-08-23
feat: add rust binding for two adic fri pcs

chokobole:
# Description

This PR adds rust binding for `TwoAdicFriPcs`.


TomTaehoonKim:
For [9c112b7](https://github.com/kroma-network/tachyon/pull/525/commits/9c112b71d6801e4b7d086bde3d732ec9a7f5a0da), 4 to 8 seems correct.

TomTaehoonKim (approved):
LGTM

tachyon/c/zk/air/plonky3/baby_bear_poseidon2_constants.h
```diff
@@ -0,0 +1,14 @@
...
+
+// Poseidon2 constants
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_WIDTH 16
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_ALPHA 7
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_FULL_ROUNDS 8
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_PARTIAL_ROUNDS 13
+
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_RATE 8
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_CHUNK 8
+#define TACHYON_PLONKY3_BABY_BEAR_POSEIDON2_N 2
```
dongchangYoo: If this config is defined by SP1, then a name like `TACHYON_PLONKY3_SP1_POSEIDON2_XX` would be more appropriate, don't you think?

dongchangYoo: If this config depends on F, then we might need to keep BABY_BEAR!

dongchangYoo (changes requested):

tachyon/c/zk/air/plonky3/baby_bear_poseidon2_constants.h
ashjeong: If you look at different cases in Plonky3, it appears that the field type does dictate different possible values for these constants as shown [here](https://github.com/Plonky3/Plonky3/blob/main/poseidon2/benches/poseidon2.rs) (note that D = ALPHA). Full rounds and partial rounds are also dependent on width, alpha, and field as stated in this comment in poseidon [here](https://github.com/Plonky3/Plonky3/blob/55832146c86e8e4d246bb9843da17f2159d212a5/poseidon/benches/poseidon.rs#L40)

vendors/sp1/src/two_adic_fri_pcs.rs
```diff
@@ -0,0 +1,123 @@
...
+            for (j, lde) in ldes.into_iter().enumerate() {
+                let v = assert_eq!(
+                    lde.to_row_major_matrix(),
+                    tachyon_data_by_round[i].ldes[j]
+                        .clone()
+                        .to_row_major_matrix()
+                );
+            }
+        }
+        // TODO(chokobole): `std::mem::forget` was used to prevent it from double-free. We need to figure out more elegant solution.
```
ashjeong, line 120:
```suggestion
        // TODO(chokobole): `std::mem::forget` was used to prevent it from double-free. We need to figure out a more elegant solution.

File: tachyon/crypto/commitments/fri/two_adic_fri_pcs_unittest.cc [hunk 3]
```diff
@@ -92,13 +91,13 @@ class TwoAdicFriPCSTest : public testing::Test {
     using Claims = std::vector<std::tuple<ExtF, std::vector<ExtF>>>;
 
     size_t num_rounds = log_degrees_by_round.size();
-    std::vector<std::vector<Domain>> domains_by_round(num_rounds);
+    std::vector<std::vector<Coset>> domains_by_round(num_rounds);
```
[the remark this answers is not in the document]
chokobole: Since the element of this vector is returned by `GetNaturalDomainForDegree()`, I left it as `domains_by_round`.

ashjeong (approved):
LGTM

batzor (approved):
LGTM

dongchangYoo (approved):
LGTM

tachyon/crypto/commitments/fri/two_adic_fri_pcs_unittest.cc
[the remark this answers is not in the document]
batzor: Actually that returns `Coset` as well.

GideokKim (approved):
LGTM

PR #531, opened 2024-09-03
perf: optimize poseidon2 pt1

ashjeong:
This PR introduces a set of intermediary commits aimed at optimizing Poseidon2. The optimizations have shown significant improvements, particularly on macOS devices. For instance, on a Mac with the M3 chip, Tachyon's runtimes have been greatly enhanced, reducing the additional time compared to Plonky3 from 60% to 30%. Note that not much change has occurred on the Ubuntu side.

To come: 
1. Refactor Poseidon/Poseidon to determine values such as `FullRounds` and `Alpha` from the start.
2. Use Sbox functions with a constexpr `Pow()`


TomTaehoonKim:
@ashjeong Can you remove scope in commit message of 5ee2c3a?

TomTaehoonKim (approved):
LGTM

benchmark/poseidon2/poseidon2_benchmark.cc
```diff
@@ -41,7 +41,6 @@ void Run(SimpleReporter& reporter, const Poseidon2Config& config, Fn horizen_fn,
     if (base::Contains(config.vendors(), Vendor::Plonky3())) {
       poseidon2_config = crypto::Poseidon2Config<Field>::CreateCustom(
           15, 7, 8, 13, math::GetPoseidon2BabyBearInternalShiftVector<15>());
-      CHECK_EQ(config.vendors().size(), static_cast<size_t>(1));
```
chokobole, line 44: Isn't it a right condition? The below constraint was if the field is `BabyBear`, the vendors should contain only `Plonky3`?

ashjeong: Yes indeed! the allowed cases are [babybear: plonky3], [babybear: horizen], and [bn254: horizen + plonky3] (implemented in this order here); however, since the constraint below was implemented, this line 44 becomes redundant as you can see in line 40, line 44 simply ensures the case for [babybear: plonky3].

chokobole:
[chore: rename functions from ...vector to ...array](https://github.com/kroma-network/tachyon/pull/531/commits/46e9bf776a6bd8565dc36de6ad1b1bc09f5a9eb5) should be typed with **refac**, I guess.
[chore(crypto): remove unneeded deps](https://github.com/kroma-network/tachyon/pull/531/commits/5ee2c3a42cb29d9c142a1f78dbce8806b3edd29f) remove scope **crypto**.
[refac(benchmark): add const to vendor loop](https://github.com/kroma-network/tachyon/pull/531/commits/a5fea22d2e7a9fbc9478265c66737f5da2aa06d0) if you search `for (Vendor`, there are more places you can change.

benchmark/poseidon2/poseidon2_config.cc
```diff
@@ -32,7 +32,7 @@ bool Poseidon2Config::Validate() const {
       tachyon_cerr << "Unsupported vendor " << vendor.ToString() << std::endl;
       return false;
     }
-    if (vendor.value() == Vendor::kPlonky3) {
+    if (prime_field_.value() == FieldType::kBabyBear) {
```
batzor, line 35: This check should be out of for loop.

chokobole (approved):
LGTM

tachyon/crypto/hashes/sponge/poseidon2/poseidon2_external_matrix.h
```diff
@@ -49,8 +49,8 @@ class Poseidon2ExternalMatrix {
       Derived::DoApply(block);
     }
 
-    math::Vector<Field> v_tmp(4);
-    v_tmp << Field::Zero(), Field::Zero(), Field::Zero(), Field::Zero();
+    std::array<Field, 4> v_tmp = {Field::Zero(), Field::Zero(), Field::Zero(),
+                                  Field::Zero()};
```
batzor, line 53: Shouldn't the function signature change as well?

`static void Apply(math::Vector<Field>& v)`

GideokKim (approved):
LGTM

dongchangYoo (approved):
LGTM

PR #539, opened 2024-09-11
feat: enable using rust allocated ldes

chokobole:
# Description

The commit to matrices are roughly performed in the order below:

`baby_bear_poseidon2::TwoAdicFri::commit()` → `tachyon_sp1_baby_bear_poseidon2_two_adic_fri_allocate_ldes()` -> `tachyon_sp1_baby_bear_poseidon2_two_adic_fri_commit()` → `FieldMerkleTreeMMCS::Commit()` → `Radix2EvaluationDomain::CosetLDEBatch()` → `ExpandInPlaceWithZeroPad()` (This generates the leaves of a merkle tree, which is a vector of the matrix(Low Degree Extension, a.k.a, LDE) which is stored in a `ProverData`).

Previously, LDE was owned by both the C++ and Rust side. In order to prevent the resulting double-free, we add an additional function `std::mem::forget()` on the Rust side. This approach forces us to modify Plonky3 code, but the problem is we can't find where to modify for `0.1.3-succinct` version in Plonky3. A naive possible solution is to allocate memory on the C++ side and de-allocate it on the Rust side, but this doesn't work because of the different memory management system and instead causes memory corruption.

To prevent a double-free or memory corruption without modifications to Plonky3 code, the Rust side should own the leaves of the `FieldMerkleTree`. The `FieldMerkleTreeMMCS::Commit()` must accept Rust allocated leaves, while the existing version is renamed to `FieldMerkleTreeMMCS::CommitOwned()` for internal use, such as unit tests. Similarly, `FieldMerkleTree::Build()` becomes `FieldMerkleTree::BuildOwned()`.

`Radix2EvaluationDomain::CosetLDEBatch()` now requires Rust allocated expanded matrices, internal memory allocation is removed, and `ExpandInPlaceWithZeroPad()` is renamed to `ExpandWithZeroPad()` .

`tachyon_sp1_baby_bear_poseidon2_two_adic_fri_commit()` is changed to accept Rust allocated leaves for the same reason.

### DIFF
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -22,25 +22,20 @@ class TwoAdicFRIImpl
   using Commitment = typename Base::Commitment;
   using ProverData = typename Base::ProverData;
   using Domain = typename Base::Domain;
-  using OpeningPoints = typename Base::OpeningPoints;
-  using OpenedValues = typename Base::OpenedValues;
-  using FRIProof = typename Base::FRIProof;
 
   using Base::Base;
 
   void AllocateLDEs(size_t size) { ldes_.reserve(size); }
 
-  template <typename Derived>
-  absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>&& matrix, F shift) {
+  void CosetLDEBatch(Eigen::Map<tachyon::math::RowMajorMatrix<F>>&& matrix,
+                     F shift,
+                     Eigen::Map<tachyon::math::RowMajorMatrix<F>>& lde) {
     Domain coset = this->GetNaturalDomainForDegree(matrix.rows());
-    tachyon::math::RowMajorMatrix<F> lde(
-        matrix.rows() << this->config_.log_blowup, matrix.cols());
     coset.domain()->CosetLDEBatch(std::move(matrix), this->config_.log_blowup,
                                   shift, lde,
                                   /*reverse_at_last=*/false);
-    absl::Span<F> ret(lde.data(), lde.size());
-    ldes_.push_back(std::move(lde));
-    return ret;
+    ldes_.push_back(Eigen::Map<const tachyon::math::RowMajorMatrix<F>>(
+        lde.data(), lde.rows(), lde.cols()));
   }
 
   using Base::Commit;
@@ -48,14 +43,13 @@ class TwoAdicFRIImpl
   void Commit(Commitment* commitment, ProverData** prover_data_out,
               std::vector<std::unique_ptr<ProverData>>* prover_data_by_round) {
     std::unique_ptr<ProverData> prover_data(new ProverData);
-    CHECK(this->mmcs_.CommitOwned(std::move(ldes_), commitment,
-                                  prover_data.get()));
+    CHECK(this->mmcs_.Commit(std::move(ldes_), commitment, prover_data.get()));
     *prover_data_out = prover_data.get();
     prover_data_by_round->push_back(std::move(prover_data));
   }
 
  protected:
-  std::vector<math::RowMajorMatrix<F>> ldes_;
+  std::vector<Eigen::Map<const tachyon::math::RowMajorMatrix<F>>> ldes_;
 };
 
 }  // namespace tachyon::c::crypto
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -33,10 +33,12 @@ class TwoAdicFRIImpl
   template <typename Derived>
   absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>& matrix, F shift) {
     Domain coset = this->GetNaturalDomainForDegree(matrix.rows());
-    tachyon::math::RowMajorMatrix<F> mat = coset.domain()->CosetLDEBatch(
-        matrix, this->fri_.log_blowup, shift, /*reverse_at_last=*/false);
-    absl::Span<F> ret(mat.data(), mat.size());
-    ldes_.push_back(std::move(mat));
+    tachyon::math::RowMajorMatrix<F> lde(matrix.rows() << this->fri_.log_blowup,
+                                         matrix.cols());
+    coset.domain()->CosetLDEBatch(matrix, this->fri_.log_blowup, shift, lde,
+                                  /*reverse_at_last=*/false);
+    absl::Span<F> ret(lde.data(), lde.size());
+    ldes_.push_back(std::move(lde));
     return ret;
   }
 
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -33,9 +33,9 @@ class TwoAdicFRIImpl
   template <typename Derived>
   absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>&& matrix, F shift) {
     Domain coset = this->GetNaturalDomainForDegree(matrix.rows());
-    tachyon::math::RowMajorMatrix<F> lde(matrix.rows() << this->fri_.log_blowup,
-                                         matrix.cols());
-    coset.domain()->CosetLDEBatch(std::move(matrix), this->fri_.log_blowup,
+    tachyon::math::RowMajorMatrix<F> lde(
+        matrix.rows() << this->config_.log_blowup, matrix.cols());
+    coset.domain()->CosetLDEBatch(std::move(matrix), this->config_.log_blowup,
                                   shift, lde,
                                   /*reverse_at_last=*/false);
     absl::Span<F> ret(lde.data(), lde.size());
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -54,25 +54,6 @@ class TwoAdicFRIImpl
     prover_data_by_round->push_back(std::move(prover_data));
   }
 
-  void CreateOpeningProof(
-      const std::vector<std::unique_ptr<ProverData>>& prover_data_by_round_in,
-      const OpeningPoints& points_by_round, Challenger& challenger,
-      OpenedValues* opened_values_by_round, FRIProof* proof) const {
-    auto& prover_data_by_round =
-        const_cast<std::vector<std::unique_ptr<ProverData>>&>(
-            prover_data_by_round_in);
-    std::vector<ProverData> prover_data_by_round_tmp = tachyon::base::Map(
-        prover_data_by_round, [](std::unique_ptr<ProverData>& prover_data) {
-          return ProverData(std::move(*prover_data));
-        });
-    CHECK(Base::CreateOpeningProof(prover_data_by_round_tmp, points_by_round,
-                                   challenger, opened_values_by_round, proof));
-    prover_data_by_round = tachyon::base::Map(
-        prover_data_by_round_tmp, [](ProverData& prover_data) {
-          return std::make_unique<ProverData>(std::move(prover_data));
-        });
-  }
-
  protected:
   std::vector<math::RowMajorMatrix<F>> ldes_;
 };
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -31,11 +31,12 @@ class TwoAdicFRIImpl
   void AllocateLDEs(size_t size) { ldes_.reserve(size); }
 
   template <typename Derived>
-  absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>& matrix, F shift) {
+  absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>&& matrix, F shift) {
     Domain coset = this->GetNaturalDomainForDegree(matrix.rows());
     tachyon::math::RowMajorMatrix<F> lde(matrix.rows() << this->fri_.log_blowup,
                                          matrix.cols());
-    coset.domain()->CosetLDEBatch(matrix, this->fri_.log_blowup, shift, lde,
+    coset.domain()->CosetLDEBatch(std::move(matrix), this->fri_.log_blowup,
+                                  shift, lde,
                                   /*reverse_at_last=*/false);
     absl::Span<F> ret(lde.data(), lde.size());
     ldes_.push_back(std::move(lde));
--- a/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
+++ b/tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
@@ -28,7 +28,7 @@ class TwoAdicFRIImpl
 
   using Base::Base;
 
-  void AllocateLDEs(size_t size) { this->ldes_.reserve(size); }
+  void AllocateLDEs(size_t size) { ldes_.reserve(size); }
 
   template <typename Derived>
   absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>& matrix, F shift) {
@@ -36,7 +36,7 @@ class TwoAdicFRIImpl
     tachyon::math::RowMajorMatrix<F> mat = coset.domain()->CosetLDEBatch(
         matrix, this->fri_.log_blowup, shift, /*reverse_at_last=*/false);
     absl::Span<F> ret(mat.data(), mat.size());
-    this->ldes_.push_back(std::move(mat));
+    ldes_.push_back(std::move(mat));
     return ret;
   }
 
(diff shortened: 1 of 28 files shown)


tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
```diff
@@ -116,7 +116,9 @@ class FieldMerkleTree {
     return {std::move(leaves), std::move(digest_layers)};
   }
 
-  const std::vector<math::RowMajorMatrix<F>>& leaves() const { return leaves_; }
+  const std::vector<Eigen::Map<math::RowMajorMatrix<F>>>& leaves() const {
+    return leaves_;
+  }
```
batzor, line 136:
```suggestion
  const std::vector<Eigen::Map<const math::RowMajorMatrix<F>>>& leaves() const {
    return leaves_;
  }
```

Otherwise, I think this is not really a `const` function.

chokobole: 1.` const Eigen::Map<math::RowMajorMatrix<F>>` will expose the methods that uses `const F*`. So I think it's fine. See https://gitlab.com/libeigen/eigen/-/blob/84282c42fc05d08216331203206de7ceb024a9c6/Eigen/src/Core/DenseCoeffsBase.h#L102-109 and https://gitlab.com/libeigen/eigen/-/blob/84282c42fc05d08216331203206de7ceb024a9c6/Eigen/src/Core/DenseCoeffsBase.h#L44-53.
2. Since `std::vector<Eigen::Map<math::RowMajorMatrix<F>>>` and `std::vector<Eigen::Map<const math::RowMajorMatrix<F>>>` are completely different type, I can't just modify the type in this way!

tachyon/crypto/commitments/fri/prove.h
```diff
@@ -56,7 +56,7 @@ CommitPhaseResult<PCS> CommitPhase(const FRIConfig<ChallengeMMCS>& config,
 
     Commitment commit;
     ProverData prover_data;
-    CHECK(config.mmcs.Commit(std::move(leaves), &commit, &prover_data));
+    CHECK(config.mmcs.CommitOwned({std::move(leaves)}, &commit, &prover_data));
```
GideokKim, line 59:
```suggestion
    CHECK(config.mmcs.CommitOwned(std::move(leaves), &commit, &prover_data));
```

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree_mmcs.h
```diff
@@ -98,6 +98,18 @@ class FieldMerkleTreeMMCS final
     return true;
   }
 
+  [[nodiscard]] bool DoCommitOwned(
+      std::vector<math::RowMajorMatrix<F>>&& owned_matrices,
+      Commitment* commitment, ProverData* prover_data) const {
+    TRACE_EVENT("ProofGeneration", "FieldMerkleTreeMMCS::DoCommit");
```
ashjeong, line 104: Can you change these to match the function name?
```suggestion
    TRACE_EVENT("ProofGeneration", "FieldMerkleTreeMMCS::DoCommitOwned");

File: tachyon/math/polynomials/univariate/radix2_evaluation_domain.h [hunk 4]
```diff
@@ -440,6 +447,31 @@ class Radix2EvaluationDomain
...
 
+  // Expands a |Eigen::MatrixBase|'s rows from |rows| to |rows|^(|added_bits|),
+  // moving values from row |i| to row |i|^(|added_bits|). All new entries are
+  // set to |F::Zero()|. Returns false when |added_bits| is zero, and no
+  // expansion is performed.
+  template <typename Derived, typename Derived2>
+  [[nodiscard]] CONSTEXPR_IF_NOT_OPENMP static bool ExpandWithZeroPad(
+      Eigen::MatrixBase<Derived>& mat, size_t added_bits,
+      Eigen::MatrixBase<Derived2>& out) {
+    if (added_bits == 0) return false;
```
ashjeong, line 455: The return value of `ExpandWithZeroPad()` seems redundant since the only place it is used already checks if `added_bits` is 0. Could you remove the return type?

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -440,6 +447,31 @@ class Radix2EvaluationDomain
     }
   }
 
+  // Expands a |Eigen::MatrixBase|'s rows from |rows| to |rows|^(|added_bits|),
+  // moving values from row |i| to row |i|^(|added_bits|). All new entries are
+  // set to |F::Zero()|. Returns false when |added_bits| is zero, and no
+  // expansion is performed.
+  template <typename Derived, typename Derived2>
```
ashjeong, line 451: It appears as though mat and out are both a `RowMajorMatrix<F>` and therefore the same type! `Derived2` seems like it can be replaced by `Derived`.

vendors/sp1/src/two_adic_fri_pcs.rs
```diff
@@ -102,7 +100,7 @@ mod test {
                     .into_iter()
                     .map(|(domain, evals)| {
                         assert_eq!(domain.size(), evals.height());
-                        let log_n = log2_strict_usize(domain.size());
+                        let _log_n = log2_strict_usize(domain.size());
```
dongchangYoo, line 123: How about removing this line?

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree_unittest.cc
```diff
@@ -178,7 +178,7 @@ TEST_F(FieldMerkleTreeTest, CommitEitherOrder) {
   Tree tree = Tree::Build(hasher_, packed_hasher_, compressor_,
                           packed_compressor_, std::move(matrices));
 
-  std::vector<math::RowMajorMatrix<F>> matrices2 = {matrix, matrix2};
+  std::vector<math::RowMajorMatrix<F>> matrices2 = {matrix2, matrix};
```
TomTaehoonKim, line 181: How did the test pass before?

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
```diff
@@ -45,16 +45,15 @@ class FieldMerkleTree {
...
+      const Hasher& hasher, const PackedHasher& packed_hasher,
+      const Compressor& compressor, const PackedCompressor& packed_compressor,
+      std::vector<Eigen::Map<math::RowMajorMatrix<F>>>&& leaves) {
     TRACE_EVENT("Utils", "FieldMerkleTree::Build");
     CHECK(!leaves.empty());
 
-    std::vector<RowMajorMatrixView> sorted_leaves =
-        base::Map(leaves, [](const math::RowMajorMatrix<F>& matrix) {
+    std::vector<RowMajorMatrixView> sorted_leaves = base::Map(
+        leaves, [](const Eigen::Map<math::RowMajorMatrix<F>>& matrix) {
```
dongchangYoo, line 56: If `Eigen::Map` has only pointer(or reference), it needs to be copy to pass to lambda function.
```suggestion
        leaves, [](Eigen::Map<math::RowMajorMatrix<F>> matrix) {
```

TomTaehoonKim:
@chokobole [4ee4d79](https://github.com/kroma-network/tachyon/pull/539/commits/4ee4d79f41cae2c0e7f29ad525dd215498473053) accept to -> to accept

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
```diff
@@ -144,23 +159,21 @@ class FieldMerkleTree {
       return ptr_->rows() < other.ptr_->rows();
     }
 
-    const math::RowMajorMatrix<F>* operator->() const { return ptr_; }
+    const Eigen::Map<math::RowMajorMatrix<F>>* operator->() const {
+      return ptr_;
+    }
 
-    const math::RowMajorMatrix<F>& operator*() const { return *ptr_; }
+    const Eigen::Map<math::RowMajorMatrix<F>>& operator*() const {
```
dongchangYoo, line 166: How about returning value by copying? That's why, it has only pointer.

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree_unittest.cc
batzor: `CommitEitherOrder` is checking if the commit result is same even if the matrix order is different. Matrix order doesn't matter I think.

chokobole: The test checks if the results remain the same for inputs containing the same elements but in a different order. Since the input elements were ordered identically, the results should also be identical.

tachyon/c/crypto/commitments/fri/two_adic_fri_impl.h
```diff
@@ -22,39 +22,32 @@ class TwoAdicFRIImpl
...
 
   using Base::Base;
 
   void AllocateLDEs(size_t size) { ldes_.reserve(size); }
 
-  template <typename Derived>
-  absl::Span<F> CosetLDEBatch(Eigen::MatrixBase<Derived>& matrix, F shift) {
+  void CosetLDEBatch(Eigen::Map<tachyon::math::RowMajorMatrix<F>>& matrix,
+                     F shift,
+                     Eigen::Map<tachyon::math::RowMajorMatrix<F>>& lde) {
```
dongchangYoo, line 32: Since `Eigen::Map` has only pointer,
```suggestion
                     Eigen::Map<tachyon::math::RowMajorMatrix<F>> lde) {
```

dongchangYoo (changes requested):

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
chokobole: There needs to be some steps to make them same type. In the commit that this change was made, the types were different. If you look at `two_adic_fri_impl.h`, `CosetLDEBatch()` used `Eigen::Map<tachyon::math::RowMajorMatrix<F>>` for `mat` and `tachyon::math::RowMajorMatrix<F>` for `out`. They became same after the commit `feat!: update c API of `CosetLDEBatch()` to accept pre-allocated memory`, so I'll make them same in this commit.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -448,10 +448,10 @@ class Radix2EvaluationDomain
   // moving values from row |i| to row |i|^(|added_bits|). All new entries are
   // set to |F::Zero()|.
   // Note that it crashes if the |added_bits| is zero.
-  template <typename Derived, typename Derived2>
+  template <typename Derived>
   CONSTEXPR_IF_NOT_OPENMP static void ExpandWithZeroPad(
       Eigen::MatrixBase<Derived>& mat, size_t added_bits,
-      Eigen::MatrixBase<Derived2>& out) {
+      Eigen::MatrixBase<Derived>& out) {
```
ashjeong, line 454: Can you change this to be in the commit it's introduced in?

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
chokobole: For the value whose size is less than or equal to 16 byte, we concluded that we should copy. But `sizeof(Eigen::Map<math::RowMajorMatrix<F>>` results in 24 byte, and I don't know if it's good to pass by value or by reference...

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -118,10 +118,10 @@ class Radix2EvaluationDomain
     ReverseMatrixIndexBits(mat);
   }
 
-  template <typename Derived>
-  CONSTEXPR_IF_NOT_OPENMP RowMajorMatrix<F> CosetLDEBatch(
+  template <typename Derived, typename Derived2>
+  CONSTEXPR_IF_NOT_OPENMP void CosetLDEBatch(
       Eigen::MatrixBase<Derived>& mat, size_t added_bits, F shift,
```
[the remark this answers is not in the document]
chokobole: I agreed, but the `mat` is changed internally.

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
```diff
@@ -161,7 +161,14 @@ class Radix2EvaluationDomain
       mat.row(base::bits::ReverseBitsLen(start + len - 1,
                                          this->log_size_of_group_)) *= weight;
     });
-    RowMajorMatrix<F> ret = ExpandInPlaceWithZeroPad(mat, added_bits);
+
+    RowMajorMatrix<F> ret;
+    if (added_bits == 0) {
+      ret = std::move(mat);
```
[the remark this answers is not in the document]
chokobole: I'll add a separate commit for this. Copy should be performed in order to preserve the logic. However, when you consider the fact that `mat` is changed, the input is difficult to be used again. Therefore, it's safe to be moved here.

dongchangYoo (approved):
LGTM

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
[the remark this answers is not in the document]
batzor: I think we can take it as value in this case. Since the `mat` will be garbage anyways, taking reference doesn't make sense. We can move if we don't need the matrix or just give it as value and it will be copied.

TomTaehoonKim (approved):
LGTM

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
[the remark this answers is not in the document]
ashjeong: Correct me if I'm wrong, but from what I've seen it doesn't seem like the matrices that are inputted into `CosetLDEBatch()` are used afterward, so perhaps moving mat (&&) would be better 🤔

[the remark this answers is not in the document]
chokobole: But then we need to be very careful when calling `CosetLDEBatch()` with this approach.
```
CosetLDEBatch(Eigen::MatrixBase<Derived>& mat,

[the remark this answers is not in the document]
Comment (batzor) [reply tachyon/math/polynomials/univariate/radix2_evaluation_domain.h:123]: If we use && we can't use const lvalue reference which is for example needed in FRI benchmark. If we just take it as value, we can move when we don't need the matrix afterwards.

```cpp
CosetLDEBatch(Eigen::MatrixBase<Derived> mat,...
// possible
CosetLDEBatch(input_ref,...);
// possible
CosetLDEBatch(std::move(input_mat...),...
```

GideokKim (approved):
LGTM

tachyon/math/polynomials/univariate/radix2_evaluation_domain.h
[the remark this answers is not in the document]
batzor: I'm not sure how that's safer than the below code. Could you  explain?

```cpp
void CosetLDEBatch(Eigen::MatrixBase<Derived> mat,...)
```

tachyon/math/polynomials/univariate/radix2_evaluation_domain_unittest.cc
ashjeong, line 1: (opt) line 53 - you can resolve if unneeded
```diff
-  RowMajorMatrix<F> matrix = RowMajorMatrix<F>::Random(1, 3);
+  const RowMajorMatrix<F> matrix = RowMajorMatrix<F>::Random(1, 3);

**Review by batzor — APPROVED** [review]
LGTM

File: tachyon/math/polynomials/univariate/radix2_evaluation_domain_unittest.cc [hunk 14]
```diff
@@ -53,22 +53,21 @@ TYPED_TEST(Radix2EvaluationDomainTest, CosetLDEBatchDeath) {
...
-    RowMajorMatrix<F> input_clone = input;
     NaiveBatchFFT<F> naive;
     F shift = F::FromMontgomery(F::Config::kSubgroupGenerator);
     RowMajorMatrix<F> expected(size_t{1} << (log_r + 1), 3);
     naive.CosetLDEBatch(input, 1, shift, expected);
     std::unique_ptr<Radix2EvaluationDomain<F>> domain =
         Radix2EvaluationDomain<F>::Create(size_t{1} << log_r);
     RowMajorMatrix<F> result(size_t{1} << (log_r + 1), 3);
-    domain->CosetLDEBatch(input_clone, 1, shift, result);
+    domain->CosetLDEBatch(input, 1, shift, result);
```
ashjeong, line 70:
```suggestion
    domain->CosetLDEBatch(std::move(input), 1, shift, result);

ashjeong (approved):
LGTM

PR #549, opened 2024-10-16
perf: add optimization flags

batzor:
# Description

In this PR, we add some compiler flags for improved performance at the cost of portability.

PR #550, opened 2024-10-16
chore: update benchmarks

ashjeong:
This PR updates all benchmarks as new flags were added from #549 and a dynamic scheduler was introduced([See here](https://github.com/kroma-network/tachyon/compare/main...dynamic)). 

The most significant decrease in time occurred for FRI on Linux. FRI on Tachyon has changed from being 1.23x faster than Plonky3 to now being 1.57x faster.

Many previous benchmarks were not updated after significant changes to its benchmark or code logic. FRI benchmark for Mac is one such example, where the `round_num` flag of 4 was added, but never re-benchmarked for Mac. As such, similar signifcant increases of time for all parties in the new benchmark data may indicate the previous benchmark was deprecated.


chokobole:
Could you change `chore(benchmark)` to `docs(benchmark)`?

chokobole (approved):
LGTM

TomTaehoonKim (approved):
LGTM

batzor (approved):
LGTM

GideokKim (approved):
LGTM

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
```diff
@@ -271,30 +271,24 @@ class FieldMerkleTree {
...
-              }
-            }
-          } else {
-            for (size_t i = 0; i < chunk.size(); ++i) {
-              chunk[i] = hasher.Hash(
-                  GetRowAsPrimeFieldVector(tallest_matrices, start + i));
-            }
-          }
-        });
+    size_t end = (sub_ret.size() / PackedPrimeField::N) * PackedPrimeField::N;
```
dongchangYoo, line 274: This repeated modification could be minimized by implementing `base::DynamicParallelizeByChunkSize`. Additionally, this function seems like it could be used continuously in the future.

dongchangYoo (changes requested):

tachyon/crypto/commitments/merkle_tree/field_merkle_tree/field_merkle_tree.h
ashjeong: I've created `base::DynamicParallelizeByChunkSize`, yet thinking about it, I'm confused on why this isn't the default for `OMP_PARALLEL_FOR` since it seems like every use case of `OMP_PARALLEL_FOR` must suffer from the same static scheduler. 🤔

dongchangYoo (approved):
LGTM

chokobole (approved):
LGTM

PR #557, opened 2024-11-22
fix(circom): delete old parts before splitting

batzor:
# Description

When compiling witness generator, sometimes # of parts becomes smaller which results in trying to compile non-overwritten old circuit parts with new parts, causing conflict. This PR fixes it by removing old split parts at the start of every run.


chokobole (approved):
LGTM

dongchangYoo (approved):
LGTM

ashjeong (approved):
LGTM

